Issue #3, opened 2020-10-21
Setup the project


BOLT04:
There has been some discussion about the technology choices for this project here: #1. So I was thinking we could get the project started and continue from there 😃.

The goal of this issue would be to bootstrap the project and get most stuff setup. So here are the tasks I purpose we complete:
- [ ] Setup a linter (e.g. ESLint)
- [ ] Setup Node.js w/TypeScript and Express
- [ ] Setup cucumber
- [ ] Setup MongoDB

eddiejaoude:
Thanks @BOLT04 . This would be good. 👍

sayand0122:
I read the issues of this project and i really feel its interesting so can someone give me ome insights on the things going on or completed?

eddiejaoude:
There are 2 PRs already open for this:

- #4 setting up the project from scratch
- #5 a spike to see if `feathersjs` is a possible option

Both are still waiting for feedback from the community 🤓

eddiejaoude:
We have broken this down into more specific issues, so I will close this.

PR #4, opened 2020-10-24
chore: initial config

AllanRegush:
completes task 1 in #3


eddiejaoude:
Looks good @AllanRegush 👍 .

I created a PR for a spike using FeathersJS. Some of the TS config is part of that, and the rest can be ported over (if we use feathersjs) - let me know what you think of feathersjs https://github.com/EddieJaoudeCommunity/EddieCommunityApi/pull/5

AllanRegush:
closing. I don't think we will use this since we are going in a different direction.

eddiejaoude:
Sorry @AllanRegush . Yep, we are going with NestJS as it looks really complete and full featured

PR #5, opened 2020-10-25
Featherjs spike

eddiejaoude:
- [x] Create featherjs app using their cli
- [x] Mongo integration with Mongoose ORM
- [x] Discord API (post, get etc)
- [x] Automated tests
- [x] [Faker](https://github.com/marak/Faker.js/) to create random test data
- [x] Code coverage
- [ ] Protect `POST` data for API to API calls, we could use [anonymous JWT](https://docs.feathersjs.com/cookbook/authentication/stateless.html)
- [ ] Swagger docs

**Example usage**

POST `localhost:3030/discord`
```json
{
    "username": "test 1"
}
```

GET `localhost:3030/discord`
```
{
    "total": 2,
    "limit": 10,
    "skip": 0,
    "data": [
        {
            "_id": "5f955b6fe839fefa05a66418",
            "username": "test 1",
            "createdAt": "2020-10-25T11:03:11.903Z",
            "updatedAt": "2020-10-25T11:03:11.903Z",
            "__v": 0
        },
        {
            "_id": "5f955b75e839fefa05a66419",
            "username": "test 2",
            "createdAt": "2020-10-25T11:03:17.301Z",
            "updatedAt": "2020-10-25T11:03:17.301Z",
            "__v": 0
        }
    ]
}
```

Screenshot 2020-10-26 at 00 22 06


AllanRegush:
I've looked over this PR. I've been thinking about the beginner developer. I feel like feathers does too much magic. Much is hidden away in the framework. Some might not know what is going on. I wonder if Express is a better option, there isn't as much done for you. Possibly, it is easier to follow along and we have alot of Python developers in our community - Express is similar to flask. Express has lot more support via tutorials, write ups, blogs, etc.

BOLT04:
I have not seen the whole PR, but I had some trouble figuring out where we have our routes and route handlers. Maybe I'm too accostumed to Express.js 😅, but it seemed that a route handler is the same as a service on Feathers.js, or maybe the service is like a router, then each route can be inside that service. But then I don't quite understand how we can have custom behaviour like where do we access the request query and other info. However this might be explained in the documentation, I just haven't read much of it.

Overall, I think it might be a slight learning curve for new contributors who are not aware of it, but I think the work that has already been done on this PR is valuable and I support using this framework for the project 👍

eddiejaoude:
Thank you both for the feedback 👍 .

@AllanRegush yes, there is quite a bit of magic going on, but as it uses conventions, there is less discussion / debate on how it should be done. We can learn ourselves from the documentation and point people at the documentation rather than explaining it to people over and over again.

@BOLT04 I do think there is a steeper learning curve, I still have a lot to learn a lot about this framework too. I just like the idea that it is well documented and has conventions on how to do things. We can point people to the documentation when they have questions on how to setup add a route/service etc.

I am still scratching the surface of this framework, I do like the idea of it and its features. I would like to do a bit more research on framework still.

AllanRegush (approved):
Looks Good 👍

BOLT04:
@eddiejaoude what do you think about merging this in and creating new issues for the leftover tasks 😃 ? I think this PR already adds value since it's the project's setup.

People in the community can start looking at it, learn and suggest changes 👍

eddiejaoude:
Good idea @BOLT04 , I wonder if we should add your swagger idea first, so that the PR contains a good example / base for new features. What do you think?

BOLT04:
agreed @eddiejaoude 👍

eddiejaoude:
Don't hate me everyone. But after some research, I think NestJS looks better, I am going to do a spike similar to this PR but in NestJS to compare.

BOLT04:
I didn't know about NestJS, but the docs do look good 😃. It's also in TypeScript and has OpenAPI support so that's good 👍 
I'm on board with another PR for it @eddiejaoude 👍.

eddiejaoude:
Thanks @BOLT04 I plan to do something on it this week 👍

eddiejaoude:
I am going to close this as we are going in a different direction with NestJS, it looks feature rich

PR #8, opened 2020-12-12
Added routes folder

mstanciu552:
Changed the file structure, for better usability. Added `routes` folder. 

Also added a message to show that the database is connected, for knowing if the database is set up correctly.(_non-essential_ let me know if this should be removed).


AllanRegush:
I think this PR is for Eddie's Feathers PR.

mstanciu552:
Well I kinda leaned into it after reading the documentation.
Didn't we already settle on feathersjs? :yum: Or are we still looking for a framework?

BOLT04:
I think @AllanRegush meant that we could change the target of this PR to the branch `featherjs`. That way this PR has only the `routes` changes 😃

AllanRegush:
Is there a way to change the target branch?

BOLT04:
> Is there a way to change the target branch?

@AllanRegush yes 😃. I think we need write access or be the user that opened the PR, but we have this `edit` button:
imagem

Then we can choose the base branch:
imagem

mstanciu552:
Is what I did correct.
My first time changing branches.
Thanks for the instruction and sorry if it is wrong.
😋 👍

BOLT04:
@mstanciu552 yup it's fine, thank you 👍

src/app.ts
```diff
@@ -35,8 +37,9 @@ app.use(compress());
 app.use(express.json());
 app.use(express.urlencoded({ extended: true }));
 app.use(favicon(path.join(app.get('public'), 'favicon.ico')));
-// Host the public folder
-app.use('/', express.static(app.get('public')));
```
BOLT04, line 39: is there a difference between `app.get('public')` and just `'public'`? Since you moved this to the routes folder just wanted to make sure the behavior is the same

src/routes/index.ts
```diff
@@ -0,0 +1,8 @@
+import express from '@feathersjs/express';
+
+const router = express.Router();
+
+// Add routes here
+router.use('/', express.static('public'));
```
BOLT04, line 6: shouldn't we have more routes here? I'm also trying to figure out how Feather.js works but perhaps routes and their handlers are called services, so maybe we could move that here?

What else do you think we can move to this file @mstanciu552 ?

src/mongoose.ts
```diff
@@ -5,8 +5,10 @@ import logger from './logger';
 export default function (app: Application): void {
   mongoose.connect(
     app.get('mongodb'),
-    { useCreateIndex: true, useNewUrlParser: true }
-  ).catch(err => {
+    { useCreateIndex: true, useNewUrlParser: true, useUnifiedTopology: true }
```
BOLT04, line 8: can you explain what the `useUnifiedTopology: true` option does? Not saying we shouldn't have it, just trying to figure out the reason for adding it 😃

BOLT04 (commented):
Good work @mstanciu552 👍, left some comments for us to discuss further

mstanciu552:
To tell you the truth I don't know exactly but the compiler stated that I have to add it .

It also gave me such a warning one ot two years back for `useNewUrlParser`.

It has something to do with deprecation for newer versions.

src/mongoose.ts
mstanciu552: Sorry I left the answer to this in the general comments. I didn't see it here.

src/app.ts
mstanciu552: I also think that it is the same but that is from @eddiejaoude so I left it there.
Maybe he can explain.

src/routes/index.ts
mstanciu552: We should have more routes.
The problem is that I didn't know what routes to make as I don't fully understand what the API should do.

This is just a structure for adding more routes.

src/mongoose.ts
BOLT04: ahh got it, the current server discovery and monitoring engine is going to be deprecated in version 4.x of MongoDB.
The release notes helped to understand the why: https://github.com/mongodb/node-mongodb-native/releases/tag/v3.2.1 😃

Thank you for explaining @mstanciu552, we should definitely have it 👍

mstanciu552: Thanks for letting me know.
👍

src/routes/index.ts
BOLT04: I see, then @eddiejaoude could you confirm if services are the same as routes in Express? Just trying to figure out if we'll have more routes in the future, or if we just add a new service.

eddiejaoude: I had to remind myself from the featherjs docs, I think we will register services not routes. Services each will manage their own routes. So maybe the idea is good but the naming needs changing?

For example...

**Service**
```ts
import { Application, Id, NullableId, Params, Service } from '@feathersjs/feathers';

class MyService implements Service<any> {
  async find(params: Params) {}
  async get(id: Id, params: Params) {}
  async create(data: any, params: Params) {}
  async update(id: NullableId, data: any, params: Params) {}
  async patch(id: NullableId, data: any, params: Params) {}
  async remove(id: NullableId, params: Params) {}
}

app.use('/my-service', new MyService());
```

**Register service**
```ts
const app = feathers();

// Register the message service on the Feathers application
app.use('messages', new MessageService());
```

https://docs.feathersjs.com/guides/basics/services.html#service-events

mstanciu552: Oh so this file should not have multiple routes but a single service that has multiple routes.

Alright I will change that. :+1:

mstanciu552:
I discovered that we don't need the `routes` folder as `feathers.js` handles that under the hood with `services`. 
I set up swagger and will be making another PR, while closing this one.

eddiejaoude:
Ok thanks @mstanciu552

PR #22, opened 2021-01-28
Create a new CRUD resource called discord

khattakdev:
This is just a boilerplate with `Get` Method.
I'll add CRUD operations in the next step.

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -1,9 +1,10 @@
 import { Module } from '@nestjs/common';
 import { AppController } from './app.controller';
 import { AppService } from './app.service';
+import { DiscordModule } from './discord/discord.module';
 
 @Module({
-  imports: [],
+  imports: [DiscordModule],
   controllers: [AppController],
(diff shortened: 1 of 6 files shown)


eddiejaoude:
Did the *generator* not create the whole CRUD?

khattakdev:
> Did the _generator_ not create the whole CRUD?

It doesn't create any API at all. It just creates an empty class
Something like this

**app.module.ts**
```ts
import { Controller } from '@nestjs/common';

@Controller()
export class AppController {}
```

**app.service.ts**
```ts
import { Injectable } from '@nestjs/common';

@Injectable()
export class AppService {}
```

eddiejaoude:
There is a command for the CRUD generator, that generates all the methods as well  https://docs.nestjs.com/recipes/crud-generator 

```
nest g resource
```

khattakdev:
`eslint` gives some warning but those will be fixed in the next Pull Request when I add some code in functions bodies.

PR #23, opened 2021-01-29
Create CRUD for member without DB

### LINKED ISSUE #20: Create a new CRUD resource called discord
Use NestJS CLI to create a new CRUD resource called `discord` with the generate command `resource` [docs](https://docs.nestjs.com/cli/usages#nest-generate)

This is a subtask (https://github.com/EddieJaoudeCommunity/api/issues/12)

### LINKED ISSUE #21: Create CRUD for member without DB
This is a subtask (https://github.com/EddieJaoudeCommunity/api/issues/12)
```ts
interface Member {
   username: string;
   bio: {
      description: string;
      twitter: string;
      linkedin: string;
      github: string;
   };
   updatedOn: Date;
   createdOn: Date;
}
```

khattakdev:
This is how Discord object looks like
```ts
export class CreateDiscordDto{
  id: number
   username: string;
   bio: {
      description: string;
      twitter: string;
      linkedin: string;
      github: string;
   };
   updatedOn: Date;
   createdOn: Date;
}
```
I've added `id` temporary, to track down objects for the update, single search, and deletion. Once we have a database, we can remove it, if we don't need it but for now, it's something needed.

I've tested the API(s) as well, and they are working perfectly fine.

**Get Single Record**
getOne

**Update Record**
patch

**Insert Record**
post

**Delete Record**
delete

**Get All Records**
getAll

Closes #21

### DIFF
--- a/src/discord/discord.service.ts
+++ b/src/discord/discord.service.ts
@@ -29,6 +29,29 @@ export class DiscordService {
   }
 
   update(id: number, updateDiscordDto: UpdateDiscordDto) {
+    const {
+      username,
+      bio: { description, twitter, linkedin, github },
+    } = updateDiscordDto;
+
+    const discordUser = this.findDiscord(id);
+    const updatedDiscord = { ...discordUser };
+    if (username) {
+      updatedDiscord.username = username;
+    }
+    if (description) {
+      updatedDiscord.bio.description = description;
+    }
+    if (twitter) {
+      updatedDiscord.bio.twitter = twitter;
+    }
+    if (linkedin) {
+      updatedDiscord.bio.linkedin = linkedin;
+    }
+    if (github) {
+      updatedDiscord.bio.github = github;
+    }
+
     return `This action updates a #${id} discord`;
   }
 
--- a/src/discord/discord.service.ts
+++ b/src/discord/discord.service.ts
@@ -1,15 +1,25 @@
 import { Injectable } from '@nestjs/common';
 import { CreateDiscordDto } from './dto/create-discord.dto';
+import { ReadDiscordDto } from './dto/read-discord.dto';
 import { UpdateDiscordDto } from './dto/update-discord.dto';
 
 @Injectable()
 export class DiscordService {
+  private readonly discord: ReadDiscordDto[] = [];
   create(createDiscordDto: CreateDiscordDto) {
-    return 'This action adds a new discord';
+    const discordUser = {
+      id: Date.now(),
+      username: createDiscordDto.username,
+      bio: { ...createDiscordDto.bio },
+      createdOn: new Date(Date.now()),
+      updatedOn: new Date(Date.now()),
+    };
+    this.discord.push(discordUser);
+    return 'User added successfully!';
   }
 
   findAll() {
-    return `This action returns all discord`;
+    return [...this.discord];
   }
 
   findOne(id: number) {
--- a/src/discord/discord.service.ts
+++ b/src/discord/discord.service.ts
@@ -23,8 +23,8 @@ export class DiscordService {
   }
 
   findOne(id: number) {
-    const discordUser = this.findDiscord(id);
-
+    const discordUser = this.discord.find((user) => user.id === id);
+    this.DiscordNotFound(discordUser);
     return { ...discordUser };
   }
 
@@ -34,7 +34,8 @@ export class DiscordService {
       bio: { description, twitter, linkedin, github },
     } = updateDiscordDto;
 
-    const discordUser = this.findDiscord(id);
+    const discordUser = this.discord.find((user) => user.id === id);
+    this.DiscordNotFound(discordUser);
     const updatedDiscord = { ...discordUser };
     if (username) {
       updatedDiscord.username = username;
@@ -52,22 +53,19 @@ export class DiscordService {
       updatedDiscord.bio.github = github;
     }
 
-    return `User updated successfully!`;
+    return 'User updated successfully!';
   }
 
   remove(id: number) {
     const userIndex = this.discord.findIndex((user) => user.id === id);
+    this.DiscordNotFound(userIndex);
     this.discord.splice(userIndex);
     return 'User deleted successfully!';
   }
 
-  private findDiscord(id: number): ReadDiscordDto {
-    const userIndex = this.discord.findIndex((user) => user.id === id);
-    const discordUser = this.discord[userIndex];
-    if (!discordUser) {
+  private DiscordNotFound(discord: ReadDiscordDto | number) {
+    if (!discord || discord < 0) {
       throw new NotFoundException('User Not Found');
     }
-
-    return discordUser;
   }
 }
--- a/src/discord/discord.service.ts
+++ b/src/discord/discord.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, NotFoundException } from '@nestjs/common';
 import { CreateDiscordDto } from './dto/create-discord.dto';
 import { ReadDiscordDto } from './dto/read-discord.dto';
 import { UpdateDiscordDto } from './dto/update-discord.dto';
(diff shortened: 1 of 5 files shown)


src/discord/discord.service.ts
```diff
@@ -1,26 +1,73 @@
...
+      updatedDiscord.bio.twitter = twitter;
+    }
+    if (linkedin) {
+      updatedDiscord.bio.linkedin = linkedin;
+    }
+    if (github) {
+      updatedDiscord.bio.github = github;
+    }
+
+    return `User updated successfully!`;
```
eddiejaoude, line 55: The return string on line `61` uses single quotes, it is better to be consistent with that if there are no variables in the string here

src/discord/discord.service.ts
```diff
@@ -1,26 +1,73 @@
...
 
   remove(id: number) {
-    return `This action removes a #${id} discord`;
+    const userIndex = this.discord.findIndex((user) => user.id === id);
+    this.discord.splice(userIndex);
+    return 'User deleted successfully!';
+  }
+
+  private findDiscord(id: number): ReadDiscordDto {
+    const userIndex = this.discord.findIndex((user) => user.id === id);
```
eddiejaoude, line 65: I think a `find` here would be cleaner, where it does not match the id, so it is excluded

src/discord/dto/read-discord.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { PartialType } from '@nestjs/mapped-types';
+import { CreateDiscordDto } from './create-discord.dto';
+
+export class ReadDiscordDto extends PartialType(CreateDiscordDto) {
+  id: number;
```
eddiejaoude, line 5: is `id` needed if it is created from `CreateDiscordDto`?

src/discord/discord.service.ts
```diff
@@ -1,26 +1,73 @@
...
   }
 
   findAll() {
-    return `This action returns all discord`;
+    return [...this.discord];
   }
 
   findOne(id: number) {
-    return `This action returns a #${id} discord`;
+    const discordUser = this.findDiscord(id);
```
eddiejaoude, line 26: I think this private function call could be replaced with a javascript `find`

src/discord/dto/read-discord.dto.ts
khattakdev: This line no `5` is making it necessary, `id` exist with that line too but it is optional ( `id?: number` ).

src/discord/discord.service.ts
eddiejaoude: In the future, the DB will be doing it, so I think it might be optimisation too soon

src/discord/dto/read-discord.dto.ts
eddiejaoude: I do not understand, where is it optional? It looks like it is required here...? https://github.com/EddieJaoudeCommunity/api/pull/23/files#diff-cdd95909b05df4daa83d67707b55752498545cda37904ddf279494e56aa4bedeR2

src/discord/discord.service.ts
khattakdev: > I think this private function call could be replaced with a javascript `find`

It's also doing error handling 
```ts
if (!discordUser) {
      throw new NotFoundException('User Not Found');
    }
```
Which I'll have to repeat on multiple locations.  How about I keep the function and replace `findIndex` with `find`? (Although it won't improve performance even when we have large data)

src/discord/dto/read-discord.dto.ts
khattakdev: I mean to say when we use `PartialType`, it will copy the values and make all of them optional (excluding nested).

Here, we have these properties in `create-discord.dto.ts`
```ts
  id: number;
  username: string;
  bio: {
    description?: string;
    twitter?: string;
    linkedin?: string;
    github?: string;
  };
  updatedOn: Date;
  createdOn: Date;
```

The top-level properties are necessary but in `read-discord.dto.ts`
If I just add
```ts
export class ReadDiscordDto extends PartialType(CreateDiscordDto) {}
```

This will results in `ReadDiscordDto` having the following properties
```ts
  id?: number;
  username?: string;
  bio?: {
    description?: string;
    twitter?: string;
    linkedin?: string;
    github?: string;
  };
  updatedOn?: Date;
  createdOn?: Date;
```

But since `id` is a primary key, which is required in every **API**, so we need to make it necessary. To do that, we can redefine that specific property in the `ReadDiscordDto`
```ts
export class ReadDiscordDto extends PartialType(CreateDiscordDto) {
  id: number;
}
```

Now this will result in
```ts
  id: number;
  username?: string;
  bio?: {
    description?: string;
    twitter?: string;
    linkedin?: string;
    github?: string;
  };
  updatedOn?: Date;
  createdOn?: Date;
```

---
In short, whatever property we define in that class will be compulsory. `PartialType` copy the properties and make it optional, if we need to make any property compulsory, we have to redefine it.

src/discord/discord.service.ts
eddiejaoude: the `find` is temporary, because it will be handled by the DB. We can leave this function for now if you like

src/discord/discord.service.ts
```diff
@@ -1,26 +1,71 @@
...
 
   remove(id: number) {
-    return `This action removes a #${id} discord`;
+    const userIndex = this.discord.findIndex((user) => user.id === id);
+    this.DiscordNotFound(userIndex);
+    this.discord.splice(userIndex);
+    return 'User deleted successfully!';
+  }
+
+  private DiscordNotFound(discord: ReadDiscordDto | number) {
```
eddiejaoude, line 66: This naming does not match the other conventions. I do not think a private method is needed for this, plus it causes type mixing

src/discord/discord.service.ts
```diff
@@ -1,26 +1,71 @@
...
+    if (github) {
+      updatedDiscord.bio.github = github;
+    }
+
+    return 'User updated successfully!';
   }
 
   remove(id: number) {
-    return `This action removes a #${id} discord`;
+    const userIndex = this.discord.findIndex((user) => user.id === id);
```
eddiejaoude, line 60: `find` would work better here, but still better to use es6 style

It will eventually be replaced by the DB

src/discord/discord.service.ts
khattakdev: What do you prefer over here? directly use `NotFoundException` instead of making a function for it?

src/discord/dto/read-discord.dto.ts
eddiejaoude: Ah I see, thank you for clarifying 👍

src/discord/discord.service.ts
khattakdev: > `find` would work better here.
> 
> It will eventually be replaced by the DB, but still better to use es6 style

I'm using splice in the next line. I am not sure how can we remove an item if we use `find`. How about we use `filter` instead of `findIndex` or `find` and update the whole record.

khattakdev: > > `find` would work better here.
> > It will eventually be replaced by the DB, but still better to use es6 style
> 
> I'm using `splice` in the next line. I am not sure how can we remove an item if we use `find`. How about we use `filter` instead of `findIndex` or `find` and update the whole record.

**Note:** `this.discord` is read-only at the moment, we will have to tweak that if we use `filter`

src/discord/discord.service.ts
eddiejaoude: yes for now please 👍  - so each area can use the object or the number, mixing types is not good

src/discord/discord.service.ts
eddiejaoude: sorry I meant `filter` not `find`, but yes `filter` out the one we no longer required would be cleaner

eddiejaoude (approved):
Looks good 👍

PR #26, opened 2021-02-14
End to End Testing

### LINKED ISSUE #25: Automated Testing
Automated testing end-to-end with what NestJS recommend (remove unit testing)

Subtask of https://github.com/EddieHubCommunity/api/issues/12

khattakdev:
- [x] Add Base Test Cases

closes #25

### DIFF
--- a/src/discord/discord.service.ts
+++ b/src/discord/discord.service.ts
@@ -5,7 +5,21 @@ import { UpdateDiscordDto } from './dto/update-discord.dto';
 
 @Injectable()
 export class DiscordService {
-  private discord: ReadDiscordDto[] = [];
+  private discord: ReadDiscordDto[] = [
+    // Dummy data
+    {
+      id: 1613291638892,
+      username: 'eddiejaoude',
+      bio: {
+        description: 'This is a GitHub Star ⭐',
+        twitter: 'eddiejaoude',
+        linkedin: 'eddiejaoude',
+        github: 'eddiejaoude',
+      },
+      createdOn: new Date('2021-02-14T08:33:58.892Z'),
+      updatedOn: new Date('2021-02-14T08:33:58.892Z'),
+    },
+  ];
   create(createDiscordDto: CreateDiscordDto) {
     const discordUser = {
       id: Date.now(),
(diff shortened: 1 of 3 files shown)


eddiejaoude:
I think the next step is to integrate **cucumberjs**

So this line

```ts
it('/Discord (Get All Users) should return an array of users', () => {
```

is replaced with something like this

```ts
Given(/^Discord "([^"]*)" should return an array of users$/, function (path) {
```

and can be used multiple times line so

```gherkin
Given I make a GET request to the API endpoint "/discord" as "users"
```

khattakdev:
> I think the next step is to integrate **cucumberjs**
> 
> So this line
> 
> ```ts
> it('/Discord (Get All Users) should return an array of users', () => {
> ```
> 
> is replaced with something like this
> 
> ```ts
> Given(/^Discord "([^"]*)" should return an array of users$/, function (path) {
> ```
> 
> and can be used multiple times line so
> 
> ```
> Given I make a GET request to the API endpoint "/discord" as "users"
> ```

But before that, Are we not suppose to make changes to the test cases. Right now, the expected values are hardcoded. 
I'm not sure how do we actually test API(s) but if we remove the dummy object, which we will do in the future, 4 out of 5 tests will fail.

eddiejaoude:
Yep, we will need to do that soon. At the moment we are slowly building up the testing and we still have more to do. We will get to dynamic tests after we add a DB. But before we add a DB, I am keen to have the tests working the way we want, so there are less moving parts, otherwise when things break it will be more difficult to understand why.

eddiejaoude:
I noticed the step def regex did not match the feature file, and also was not good English. This would be better...

```ts
When(/^I make a "([^"]*)" request to "([^"]*)" as "([^"]*)"$/, function (method, path, users) {
```

to match 

```gherkin
When I make a "GET" request to the API endpoint "/discord" as "users"
```

Then the next step would be something like

```gherkin
Then I should get the data ...
```

andrew-cunliffe:
Step definitions should not use pronouns.... avoid using I / he / she / it

```gherkin
When a "GET" request is made to the "/path" endpoint as "path"
Then the body of the "path" response should match....
```

eddiejaoude:
Yes good point 👍

khattakdev:
This PR is invalid, the changes have been merged via #26

PR #27, opened 2021-02-25
Cucumber E2E testing

khattakdev:
- [x] Add E2E Testing
- [x] Wrapped the tests with Cucumber


test/features/discord.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Discord Module
+
+  Scenario: Get List of Users
+    When make a GET Request to "/Discord"
```
eddiejaoude, line 4: I think this should be a `given`, for example

`Given a GET request is made to "/discord"

Note: I think `request` and `discord` should be lowercase

test/step-definitions/responses.ts
```diff
@@ -0,0 +1,49 @@
...
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    this.context.app = moduleFixture.createNestApplication();
+    await this.context.app.init();
+  }
+
+  @when(/make a GET Request to "([^"]*)"/)
+  public async callToGETAPI(url: string) {
```
eddiejaoude, line 28: `GETAPI` this in capitals makes it difficult to read.

What about the method name to be something like `getRequest`?

eddiejaoude:
Looks good 👍 , a few small comments

test/features/discord.feature
khattakdev: From the [docs](https://cucumber.io/docs/gherkin/reference/#example)

What I understand is:
**Given** is used to describe the initial context of the system, we should avoid user interaction here
**When** is used to describe an event or an action.
**Then** is used to describe an expected outcome, or result.

This `make a GET request to "/discord"` is a user action, I think it's best to have it in `When`
For Given we can have
`Given There is a (GET) API Endpoint "/discord"`

CODE_OF_CONDUCT.md
```diff
@@ -14,22 +14,22 @@ appearance, race, religion, or sexual identity and orientation.
...
-* Other conduct which could reasonably be considered inappropriate in a
- professional setting
+- The use of sexualized language or imagery and unwelcome sexual attention or
+  advances
+- Trolling, insulting/derogatory comments, and personal or political attacks
+- Public or private harassment
+- Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+- Other conduct which could reasonably be considered inappropriate in a
+  professional setting
```
eddiejaoude, line 32: Some of my thoughts on these changes

- This is from an open source template, so I think it is best to leave it it unchanged
- Also formation changes like these should be in a separate PR

package.json
```diff
@@ -18,37 +17,43 @@
...
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json"
+    "test:e2e": "cucumber-js -p default"
   },
   "dependencies": {
-    "@nestjs/common": "^7.5.1",
-    "@nestjs/core": "^7.5.1",
-    "@nestjs/platform-express": "^7.5.1",
+    "@nestjs/common": "^7.6.13",
+    "@nestjs/core": "^7.6.13",
```
eddiejaoude, line 24: Version bumps should be in its own PR in the future

CODE_OF_CONDUCT.md
khattakdev: This actually happened due to `Prettier`. I Will reset it.

test/features/discord.feature
```diff
@@ -0,0 +1,34 @@
+Feature: Discord Module
+
+  Scenario: Get List of Users
```
eddiejaoude, line 3: If everything can be lower case

eddiejaoude: keep things like `GET` upper case though

package.json
khattakdev: There was some dependencies issue, a few of the dependencies were not working with cucumber, so I had to change the versions.

test/features/discord.feature
```diff
@@ -0,0 +1,34 @@
...
+    Given make a GET request to "/discord"
+    Then the response status code should be "200"
+    And the response should be "[]"
+
+
```
eddiejaoude, line 8: Double blank lines, only 1 is needed

test/features/discord.feature
```diff
@@ -0,0 +1,34 @@
...
+  Scenario: Add a New User
+    Given make a POST request to "/discord"
+    Then the response status code should be "201"
+    And the response should be "User added successfully!"
+
+  Scenario: Add an Empty User
+    Given make a POST request without body to "/discord"
+    Then the response status code should be "400"
+    And the response should contains:
+     | statusCode | message       |
```
eddiejaoude, line 18: Table not aligned, there is a VSCode plugin to do this

test/features/discord.feature
```diff
@@ -0,0 +1,34 @@
+Feature: Discord Module
+
+  Scenario: Get List of Users
+    Given make a GET request to "/discord"
+    Then the response status code should be "200"
+    And the response should be "[]"
+
+
+  Scenario: Add a New User
+    Given make a POST request to "/discord"
```
eddiejaoude, line 10: We will need to add a table so POST data can be added/changed here, but this can come in a future PR

test/step-definitions/responses.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.context.response = await request(this.context.app.getHttpServer())
+      .put(url + `/${user.id}`)
+      .send({
+        bio: {
+          description: 'This is Campus Expert 🚩 from Pakistan',
+        },
+      });
+  }
+
+  @when(/make a DELETE request to "([^"]*)" with an ID/)
```
eddiejaoude, line 72: I think these methods above can go into a `request` file and the methods below will be in the `responses` file

CODE_OF_CONDUCT.md
eddiejaoude: oh I see, no problem you can leave it

package.json
eddiejaoude: ah ok 👍

eddiejaoude (approved):
Looks good 👍

PR #35, opened 2021-03-15
Add MongoDB for CRUD with typeorm

### LINKED ISSUE #34: Add Mongo DB for CRUD with typeorm
Currently, the data is managed (CRUD) in Memory. 
It needs to be replaced with `typeorm` using `MongoDB`

khattakdev:
- [x] Create Record in DB
- [x] Read Record in DB
- [x] Update Record in DB
- [x] Delete Record in DB
- [x] Updare README (add .env instructions)
- [x] Fix unit tests
- [ ] Fix e2e tests

closes #34


eddiejaoude:
I added the mongo service, but it still is not connecting. I suggest next step is to create a simple mongo js script to run in the action to see if we can ping it. I saw something like this

```js
var connectionString = "mongodb://localhost";
var client = new MongoClient(connectionString);
var server = client.GetServer();
server.Ping();
```

eddiejaoude:
Thank you for the PR. But this is no longer needed.

PR #36, opened 2021-03-27
Replace MongoDb with Datastax

khattakdev:
Replaced The MongoDB crud with [datstax](http://links.eddiejaoude.io/t/dsXGWMoY9uCp) for performance test


README.md
```diff
@@ -29,6 +29,14 @@ An API to manage our community data
 $ npm install
 ```
 
+Create a `.env` file in the project root folder, with the following content
+
+```
+MONGODB_CONNECTION_STRING=<Connection URI>
+MONGODB_DATABASE=<Database name>
+MONGODB_DATABASE_TEST=<Database name>
```
eddiejaoude, line 37: This will need updating to DataStax's Astra environment variables

package.json
```diff
@@ -9,28 +9,33 @@
...
-    "test:e2e": "cucumber-js -p default"
+    "test:e2e": "cross-env NODE_ENV=test cucumber-js -p default"
   },
   "dependencies": {
+    "@astrajs/collections": "0.0.12",
     "@nestjs/common": "^7.6.13",
     "@nestjs/core": "^7.6.13",
     "@nestjs/mapped-types": "^0.3.0",
     "@nestjs/platform-express": "^7.6.13",
+    "cassandra-driver": "^4.6.1",
```
eddiejaoude, line 28: Is this needed?

package.json
```diff
@@ -39,21 +44,24 @@
...
     "chai": "^4.3.0",
+    "cross-env": "^7.0.3",
     "cucumber": "^5.1.0",
     "cucumber-pretty": "^1.5.2",
     "cucumber-tsflow": "^3.4.1",
+    "eslint": "^7.12.1",
+    "eslint-config-prettier": "7.2.0",
+    "eslint-plugin-prettier": "^3.1.4",
     "jest": "26.6.3",
+    "mongodb": "^3.6.4",
```
eddiejaoude, line 56: Is this needed?

src/discord/discord.service.ts
```diff
@@ -1,75 +1,91 @@
...
 import { ReadDiscordDto } from './dto/read-discord.dto';
 import { UpdateDiscordDto } from './dto/update-discord.dto';
 
-@Injectable()
 export class DiscordService {
-  private discord: ReadDiscordDto[] = [];
-  create(createDiscordDto: ReadDiscordDto) {
+  astraClient;
+  usersCollection;
+  basePath = '/api/rest/v2/KEYSPACES/<namespace>/collections/<collectionName>';
```
eddiejaoude, line 10: Is `basePath` used?

src/discord/discord.service.ts
```diff
@@ -1,75 +1,91 @@
...
+
 import { ReadDiscordDto } from './dto/read-discord.dto';
 import { UpdateDiscordDto } from './dto/update-discord.dto';
 
-@Injectable()
 export class DiscordService {
-  private discord: ReadDiscordDto[] = [];
-  create(createDiscordDto: ReadDiscordDto) {
+  astraClient;
+  usersCollection;
```
eddiejaoude, line 9: Could `astraClient` and `usersCollection` be added to the constructor parameters which then are automatically set as class variables

```ts
constructor(private astraClient: AstraClient, private usersCollection: Collection)
```

Note: check the types, I just took a guess here as an example

khattakdev: `@astrajs` is not really designed for TypeScript, so it doesn't have types defined for `astraClient` and `userCollection`. We have to go with any or looking at the returning value of `astraClient` and `userCollection` we need to create types, which I am not sure if they will have the same values all the time or might changes in some scenarios.

@eddiejaoude  Need your suggestion on this.

eddiejaoude: I would us these types for now. It is better than `any`

src/discord/discord.service.ts
```diff
@@ -1,75 +1,91 @@
...
+  basePath = '/api/rest/v2/KEYSPACES/<namespace>/collections/<collectionName>';
+  constructor() {
+    (async () => {
+      try {
+        this.astraClient = await createClient({
```
eddiejaoude, line 14: I think this should be moved to a provider and use NestJS config

- config https://docs.nestjs.com/techniques/configuration
- provider https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory

src/discord/discord.service.ts
```diff
@@ -1,75 +1,91 @@
...
+    await this.usersCollection.create({
+      ...discordUser,
+    });
+    return 'User added successfully! ';
   }
 
-  findAll() {
-    return [...this.discord];
+  async findAll() {
+    return await this.usersCollection.find();
```
eddiejaoude, line 47: I think NestJS resolves the promise in the `controller` and from the service a pro use or observable should be returned - maybe we be consistent and use observables

src/app.module.ts
```diff
@@ -2,9 +2,9 @@ import { Module } from '@nestjs/common';
 import { AppController } from './app.controller';
 import { AppService } from './app.service';
 import { DiscordModule } from './discord/discord.module';
-
+import { ConfigModule } from '@nestjs/config';
 @Module({
-  imports: [DiscordModule],
+  imports: [ConfigModule.forRoot(), DiscordModule],
```
Cahllagerfeld, line 7: In case we need ConfigModule in other modules as well theres the posibility to import it as a global Module by giving the options-object with the following configuration
```ts
{
  isGlobal: true,
}
``` 
to the `forRoot` function. 
In this case when we're just having a single module this is fine I guess, but as the application grows we can save this import in any others module then and just have to do it once.

eddiejaoude:
I think this is out of date now, so I will close

Issue #44, opened 2021-04-12
GitHub data collection


eddiejaoude:
**GitHub**

```ts
interface GitHub {
   username: string;
   bio: string; // do we need Discord bio?
   avatarUrl: string
   followers: number;
   repos: number;
   // ... any other useful info public info https://docs.github.com/en/rest/reference/users
   communityStats: {
      [key: string]: number
   };
}
```

Also playing with https://github.com/EddieHubCommunity/api/issues/61

Cahllagerfeld:
Is this Issue meant as a Github-Integration within the Nest-App?

eddiejaoude:
Oh sorry just seeing this notification now. We have a GitHub Action that will send stats from our community Org Repos to our API

Cahllagerfeld:
Thats nice! Luckily we dont have to handle different Accesstokens for different users then.

Cahllagerfeld:
I'll start working on this. You can assign this Issue to me if you want.

eddiejaoude:
Great, thanks 👍 . Sure assigned 🎉

Cahllagerfeld:
> 
> 
> **GitHub**
> 
> ```ts
> interface GitHub {
>    username: string;
>    bio: string; // do we need Discord bio?
>    avatarUrl: string
>    followers: number;
>    repos: number;
>    // ... any other useful info public info https://docs.github.com/en/rest/reference/users
>    communityStats: {
>       [key: string]: number
>    };
> }
> ```
> 
> Also playing with #61

Hi Eddie, could you please specify which of the properties are mandatory, and which of them are optional?
Thank you! :D

eddiejaoude:
Thanks for the questions @Cahllagerfeld , here are my thoughts...

> could you please specify which of the properties are mandatory, and which of them are optional?

I think the `username` is the only required property

> Will we accept any Key, Value pairs in communityStats Object, or do we need to specify this more?

Ideally we should specify these, and do a check that it is in our allowed list. The list will come from GitHub events, here is the [example usage](https://github.com/EddieHubCommunity/gh-action-community/blob/develop/src/statistics/index.js#L16) for where the data will come from, and here are the [docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows) for the list of events

 > Should we implement the Validationpipe mentionend in #41?

I think that would be great 👍 , it would allow us to re-use the validation in multiple places

Cahllagerfeld:
[...]

PR #54, opened 2021-04-13
Fix Issue #11 - Standup Module

### LINKED ISSUE #11: Standup endpoint
It would be great to encourage more daily standup information. 

- Save (from bot or UI) daily standup information
- Read daily standup information in daily order desc

---

Update after discussion with @mstanciu552 below

> send the information from EddieBot to the API using axios or fetch and then save them in a users collection in our db

Cahllagerfeld:
closes #11 

- [x] Initialize Endpoint for Standups
- [x] Updated Swagger-Documentation

Screen
- [ ] Currently only SearchContext is Discord-Username; This could be extended
- [ ] An Array is used to store date in (like in `discord-service`). This needs to be changed as soon as a Databse is available

### DIFF
--- a/src/standup/interfaces/Standup.interface.ts
+++ b/src/standup/interfaces/Standup.interface.ts
@@ -1,0 +1,7 @@
+export interface IStandup {
+  id: number;
+  yesterdayMessage: string;
+  todayMessage: string;
+  discordUser: string;
+  createdOn: Date;
+}
--- a/src/standup/interfaces/Standup.interface.ts
+++ b/src/standup/interfaces/Standup.interface.ts
@@ -1,4 +1,4 @@
-export interface IStandup {
+export interface Standup {
   id: number;
   yesterdayMessage: string;
   todayMessage: string;
--- a/src/standup/standup.controller.ts
+++ b/src/standup/standup.controller.ts
@@ -1,0 +1,45 @@
+import {
+  Body,
+  Controller,
+  Delete,
+  Get,
+  Param,
+  ParseIntPipe,
+  Post,
+  Query,
+} from '@nestjs/common';
+import { ApiTags } from '@nestjs/swagger';
+import { CreateStandupDTO } from './dto/CreateStandup.dto';
+import { SearchStandupDTO } from './dto/SearchStandup.dto';
+import { StandupService } from './standup.service';
+
+@ApiTags('Standup')
+@Controller('standup')
+export class StandupController {
+  constructor(private readonly standupService: StandupService) {}
+
+  @Post()
+  createStandup(@Body() body: CreateStandupDTO) {
+    return this.standupService.create(body);
+  }
+
+  @Get()
+  findAllStandups() {
+    return this.standupService.findAll();
+  }
+
+  @Get(':id')
+  findById(@Param('id', new ParseIntPipe()) id) {
+    return this.standupService.findById(id);
+  }
+
+  @Delete(':id')
+  deleteStandup(@Param('id', new ParseIntPipe()) id: number) {
+    return this.standupService.deleteStandup(id);
+  }
+
+  @Post('search')
+  search(@Body() body: SearchStandupDTO) {
+    return this.standupService.search(body);
+  }
+}
--- a/src/standup/standup.controller.ts
+++ b/src/standup/standup.controller.ts
@@ -6,7 +6,6 @@ import {
   Param,
   ParseIntPipe,
   Post,
-  Query,
 } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { CreateStandupDTO } from './dto/CreateStandup.dto';
--- a/src/standup/standup.controller.ts
+++ b/src/standup/standup.controller.ts
@@ -26,6 +26,7 @@ export class StandupController {
   findAllStandups() {
     return this.standupService.findAll();
   }
+
   @Get('search')
   @ApiQuery({ name: 'discordUser', type: 'string' })
   search(@Query() query) {
--- a/src/standup/standup.controller.ts
+++ b/src/standup/standup.controller.ts
@@ -6,10 +6,10 @@ import {
   Param,
   ParseIntPipe,
   Post,
+  Query,
 } from '@nestjs/common';
-import { ApiTags } from '@nestjs/swagger';
+import { ApiQuery, ApiTags } from '@nestjs/swagger';
 import { CreateStandupDTO } from './dto/CreateStandup.dto';
-import { SearchStandupDTO } from './dto/SearchStandup.dto';
 import { StandupService } from './standup.service';
 
 @ApiTags('Standup')
@@ -26,6 +26,11 @@ export class StandupController {
   findAllStandups() {
     return this.standupService.findAll();
   }
+  @Get('search')
+  @ApiQuery({ name: 'discordUser', type: 'string' })
+  search(@Query() query) {
+    return this.standupService.search(query);
+  }
 
   @Get(':id')
   findById(@Param('id', new ParseIntPipe()) id) {
@@ -35,10 +40,5 @@ export class StandupController {
   @Delete(':id')
   deleteStandup(@Param('id', new ParseIntPipe()) id: number) {
     return this.standupService.deleteStandup(id);
-  }
-
-  @Post('search')
-  search(@Body() body: SearchStandupDTO) {
-    return this.standupService.search(body);
   }
 }
--- a/test/features/standup.feature
+++ b/test/features/standup.feature
@@ -19,7 +19,6 @@ Feature: Standup-module
             | id               | 666                        |
 
 
-    //TODO Step Definition Array of Objects
     Scenario: search Standup
         Given make a POST request to "/standup" with:
             | discordUser      | "eddiehubber"          |
@@ -28,7 +27,12 @@ Feature: Standup-module
         Then  make a POST request to "/standup/search" with:
             | discordUser | "eddiehubber" |
         Then the response status code should be 201
-        And  the response should be:
+        And  the response in item "0" should contains:
+            | discordUser      | "eddiehubber"              |
+            | yesterdayMessage | "yesterday I did this"     |
+            | todayMessage     | "Today I'll do this"       |
+            | createdOn        | "2021-01-01T00:00:00.000Z" |
+            | id               | 666                        |
 
 
     Scenario: add an empty standup
--- a/test/features/standup.feature
+++ b/test/features/standup.feature
@@ -1,0 +1,49 @@
+Feature: Standup-module
+
+    Scenario: get list of standups
+        Given make a GET request to "/standup"
+        Then the response status code should be 200
+        And the response should be "[]"
+
+    Scenario: add a new standup
+        Given make a POST request to "/standup" with:
+            | discordUser      | "eddiehubber"          |
+            | yesterdayMessage | "yesterday I did this" |
+            | todayMessage     | "Today I'll do this"   |
+        Then the response status code should be 201
+        And the response should contains:
+            | discordUser      | "eddiehubber"              |
+            | yesterdayMessage | "yesterday I did this"     |
+            | todayMessage     | "Today I'll do this"       |
+            | createdOn        | "2021-01-01T00:00:00.000Z" |
+            | id               | 666                        |
+
+
+    //TODO Step Definition Array of Objects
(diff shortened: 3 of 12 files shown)


eddiejaoude:
Thank you 👍 . I will review after we merge #49 so we can also add automated tests. Then when we connect a database, the tests should still pass 🤓

src/standup/interfaces/Standup.interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface IStandup {
```
eddiejaoude, line 1: Is this the recommended way to do interface naming in NestJS? From typescript no `I` prepending is needed https://www.typescriptlang.org/docs/handbook/2/objects.html

Cahllagerfeld: To be honest I don't really know.. A colleague at work always did this, so I adapted my style to that.
I could change this if its not necessary.

eddiejaoude: I like to keep things vanilla / recommended from official docs, because it makes it easier to onboard new people. If you could please update 👍

Cahllagerfeld:
Hi @eddiejaoude,

the checks are passing locally at my Computer, I cant really explain why they fail here..
I dont know what that these `fsevents` are..
You have any suggestion what to do?

eddiejaoude:
No problem, I can look into the issues soon.

Btw we also write end-to-end tests, here is an example https://github.com/EddieHubCommunity/api/blob/main/test/features/discord.feature any questions let us know

Cahllagerfeld:
> 
> 
> Hi @eddiejaoude,
> 
> the checks are passing locally at my Computer, I cant really explain why they fail here..
> I dont know what that these `fsevents` are, but they seem to be the reason for failure
> You have any suggestion what to do?

The Problem when passing the checks seems to be while installing the dependencies, not while the actual check itself.

 
> 
> No problem, I can look into the issues soon.
> 
> Btw we also write end-to-end tests, here is an example https://github.com/EddieHubCommunity/api/blob/main/test/features/discord.feature any questions let us know

Sure I'll add some.
I'll probably wont have enough time today to initialize and commit the e2e-Tests today. For this reason I'll suggest leaving this PR open until I updated them.

eddiejaoude:
> The Problem when passing the checks seems to be while installing the dependencies, not while the actual check itself.

it might be a version incompatibility issue. What version of node are you using locally? I can test locally also later on.

> I'll probably wont have enough time today to initialize and commit the e2e-Tests today. For this reason I'll suggest leaving this PR open until I updated them.

No problem, we can leave it open 👍

Cahllagerfeld:
My local Node-Version is `v14.15.4`
Local npm is `7.9.0`

Cahllagerfeld:
Hi @eddiejaoude as discussed on the discord server, we should introduce step-definitions for responses. In this case we need a step-definition for testing the `standup-search`-endpoint. The datastructure is the following: 
```ts 
[
  {
    "id": number,
    "yesterdayMessage": string,
    "todayMessage": string,
    "discordUser": string,
    "createdOn": Date
  }
]
```

You prefer collborating directly in this PR here, or open an issue for that?

eddiejaoude:
Thanks @Cahllagerfeld for the context 👍 

I think we are missing this functionality. 
We need a step definition like this (untested) in file `test/step-definitions/responses.ts` 
```ts
@then(/the response in item "([^"]*)" should contains:/)
public dataResponseItemTable(item: number, table: { rawTable: [] }) {
  const data = this.context.tableToObject(table);
  expect(JSON.parse(this.context.response.text)[item]).to.to.eql(data);
}
```

test/features/standup.feature
```diff
@@ -0,0 +1,49 @@
...
+        Then the response status code should be 201
+        And the response should contains:
+            | discordUser      | "eddiehubber"              |
+            | yesterdayMessage | "yesterday I did this"     |
+            | todayMessage     | "Today I'll do this"       |
+            | createdOn        | "2021-01-01T00:00:00.000Z" |
+            | id               | 666                        |
+
+
+    //TODO Step Definition Array of Objects
```
Cahllagerfeld, line 22: Thats the Test for the Search Endpoint, which returns an Array ob Standup-Objects.

test/features/standup.feature
```diff
@@ -0,0 +1,49 @@
...
+    Scenario: search Standup
+        Given make a POST request to "/standup" with:
+            | discordUser      | "eddiehubber"          |
+            | yesterdayMessage | "yesterday I did this" |
+            | todayMessage     | "Today I'll do this"   |
+        Then  make a POST request to "/standup/search" with:
+            | discordUser | "eddiehubber" |
+        Then the response status code should be 201
+        And  the response should be:
```
eddiejaoude, line 31: Is this missing the return data?

Cahllagerfeld: Exactly, I didnt know how to project the Data-Structure described earlier in this PR as a Datatable that cucumber can validate it. Because of that reason I dont have a check for the Response yet

test/features/standup.feature
```diff
@@ -0,0 +1,53 @@
...
+
+    Scenario: search Standup
+        Given make a POST request to "/standup" with:
+            | discordUser      | "eddiehubber"          |
+            | yesterdayMessage | "yesterday I did this" |
+            | todayMessage     | "Today I'll do this"   |
+        Then  make a POST request to "/standup/search" with:
+            | discordUser | "eddiehubber" |
+        Then the response status code should be 201
+        And  the response in item "0" should contains:
```
Cahllagerfeld, line 30: This line uses the new Step-Definition. I am not sure if this is how it is intended to work, but locally it passed the test successfully with this.

Cahllagerfeld:
I reinvestigated the failure of the checks. It seems like it has sth to do with this commit. `c22711238127eff1543510f4038fb2a2dd9ba1c5`. In this Commit about 11000 lines of code get deleted from package-lock. When I delete `package-lock` locally and reinstall dependencies these 11000 lines are getting added back. Maybe - no guarantee - this causes the checks to fail.

eddiejaoude:
Thank you for making the changes 🤓 

I think you might be right, try deleting the lock file and doing an `npm install` and committing the new lock file 👍

I ran the tests locally and they pass 🎉 , I also purposely broke the test to make sure it was testing and it failed correctly (see screenshot below)

This is vey close to be able to merge, thank you 💥 

---

Testing a broken test fails...

Screenshot 2021-04-16 at 09 15 29

test/features/standup.feature
eddiejaoude: ok thanks 👍

test/features/standup.feature
eddiejaoude: ok great 👍

test/features/standup.feature
```diff
@@ -0,0 +1,53 @@
...
+            | createdOn        | "2021-01-01T00:00:00.000Z" |
+            | id               | 666                        |
+
+
+    Scenario: search Standup
+        Given make a POST request to "/standup" with:
+            | discordUser      | "eddiehubber"          |
+            | yesterdayMessage | "yesterday I did this" |
+            | todayMessage     | "Today I'll do this"   |
+        Then  make a POST request to "/standup/search" with:
```
eddiejaoude, line 27: I think `search` should be a `GET` with a query parameter, something like `GET /standup?discordUser=eddiehubber`

Cahllagerfeld: I had the same thoughts while coding this. I made it a Post in case the SearchContext is getting more complex in the future so the URL doesnt get too long with too many parameters. 
Do we need more properties be searchable in this context when I adjust this endpoint?

eddiejaoude: I think keep it simple for now.

In the future we might want
- keyword search that searches `yesterdayMessage` and `todayMessage`
- date range filter (for example list the last 3 days for all users, or filter by user as well)

Cahllagerfeld: Okay I'll implement the simple way for now then! :D

src/standup/standup.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+  @Post()
+  createStandup(@Body() body: CreateStandupDTO) {
+    return this.standupService.create(body);
+  }
+
+  @Get()
+  findAllStandups() {
+    return this.standupService.findAll();
+  }
+  @Get('search')
```
Cahllagerfeld, line 29: Search is now get

test/features/standup.feature
```diff
@@ -19,21 +19,41 @@ Feature: Standup-module
...
+        Then  make a GET request to "/standup/search?discordUser=eddiehubber"
+        Then the response status code should be 200
         And  the response in item "0" should contains:
             | discordUser      | "eddiehubber"              |
             | yesterdayMessage | "yesterday I did this"     |
             | todayMessage     | "Today I'll do this"       |
             | createdOn        | "2021-01-01T00:00:00.000Z" |
             | id               | 666                        |
 
+    Scenario: search non-existing Standup
```
Cahllagerfeld, line 36: Added some additional Tests

src/standup/standup.controller.ts
eddiejaoude: Great thanks 👍 

Btw I think a blank line is needed between the methods

test/features/standup.feature
```diff
@@ -0,0 +1,85 @@
...
+            | todayMessage     | "Today I'll do this"   |
+        Then  make a GET request to "/standup/search"
+        Then the response status code should be 400
+        And  the response should contains:
+            | statusCode | 400                             |
+            | message    | "Please Provice Search Context" |
+
+
+
+    Scenario: add an empty standup
```
eddiejaoude, line 58: `standup` here is lowercase and in other places it has a capital. We should try to be consistent. I think having everything lower case in the `Scenario` would make it easier for everyone to be consistent

For example https://github.com/EddieHubCommunity/api/pull/54/files#diff-757cd4aa33cb116bb7e4ebeef9a619042a9b5f4fe6acae2b75bdfbba93b74ca1R45

eddiejaoude:
Looks great 👍 , I left some small comments.

test/features/standup.feature
Cahllagerfeld: Yes, I definitly see this issue. 
I'll fix this.

Cahllagerfeld:
I hope everything is fine now :D. Now the only thing to fix is the problem with the checks :grimacing:

Cahllagerfeld:
> 
> 
> I hope everything is fine now :D. Now the only thing to fix is the problem with the checks 😬

Hi Eddie, i did some more research on that. Did you do `npm i` or `npm ci` after pull? The difference is, that `npm ci` doesnt write to `package-lock.json`. 
I did `npm i` when I installed the dependencies. This means that I potentially overwrote some entries in the `package-lock.json`. 
You think restoring the lockfile from [this](https://github.com/EddieHubCommunity/api/commit/c22711238127eff1543510f4038fb2a2dd9ba1c5) commit might fix the problem?

eddiejaoude:
> I did npm i when I installed the dependencies. This means that I potentially overwrote some entries in the package-lock.json

I think you should be able to create a new lock file, which should be ok. Let me test locally also

eddiejaoude:
btw I tested it locally and it works fine for me. I think it is still a good idea to generate a new lock file though

I have node `v14.14.0` locally

eddiejaoude:
I will merge this into a temporary branch `issue-11` on the repo, so I can have a play with changes too 🤓 - feel free to create more PR against `issue-11` if you would like to make any changes to this issue also

Cahllagerfeld:
Thats a good idea!

eddiejaoude:
That seems to have fixed it 🎉 #62

Cahllagerfeld:
Thats huge!! :rocket:
Which issues do you see next to be implemented?
I feel quite motivated to grow the api, but I cant really prioritize the open issues and cant decide on which one to work next. Could you help me about that? :sweat_smile:

eddiejaoude:
Your code is now in the `main` branch 🎉 - thank you 🤓 

Wow thank you for your awesome support! Yes the issues are not organised at the moment and that is my fault 🤦‍♂️ 

Could you add a GitHub module, similar to the other modules with tests and swagger etc https://github.com/EddieHubCommunity/api/issues/44

Any questions let us know 🤓

Cahllagerfeld:
Sure :D Then I'll start the Github-Module. I open a new PR then.

Issue #38, opened 2021-04-12
Discord data collection


eddiejaoude:
**Discord**
```ts
interface Discord {
   uid: string;
   location: {
       text: string;
       lat: number;
       lng: number;
   };
   socials: {
       [key: string]: string; // GitHub, linkedin, instagram, twitter
   }
}
```

Cahllagerfeld:
I think except the location we're having this module already implemented.
Are we going to need a Geocoding Service to turn real places into coordinates or can we expect the Client to send them?

eddiejaoude:
Great question, we need to get the lat/lng, here is an example where we have done this before, but it will need moving to this project and module https://github.com/EddieHubCommunity/EddieBot/blob/ed49a8a3dd55d23b0e1d92bd09f2ea4f673993d8/src/commandHandlers/bio.ts#L128-L132

Cahllagerfeld:
Okay I see this logic.
I suggest using the [nest-http-module](https://docs.nestjs.com/techniques/http-module). It is axios-based and uses observables out-of-the-box-

eddiejaoude:
Sounds like a good idea 👍 , I like to keep things vanilla and go on the official recommended of the framework

PR #63, opened 2021-04-17
Issue 44 - Github Data Collection

### LINKED ISSUE #44: GitHub data collection

Cahllagerfeld:
closes #44 
Module which collects Github Data which are sent from an Action.

### DIFF
--- a/src/github/communitystats-mapping.service.ts
+++ b/src/github/communitystats-mapping.service.ts
@@ -47,9 +47,9 @@ export class CommunitystatsMappingService {
 
   public mapCommunityState(
     event: string,
-    existingStats: CommunitystatsMappingService | {},
+    existingStats: CommunityStats,
   ): CommunityStats {
-    let stats = { ...existingStats };
+    const stats = { ...existingStats };
     const mappedValue = this.mapEvent(event);
     if (mappedValue in stats) {
       stats[mappedValue] += 1;
--- a/src/github/communitystats-mapping.service.ts
+++ b/src/github/communitystats-mapping.service.ts
@@ -1,0 +1,46 @@
+import { Injectable, BadRequestException } from '@nestjs/common';
+
+@Injectable()
+export class CommunitystatsMappingService {
+  private eventMap = {
+    workflow_dispatch: 'workflowDispatch',
+    repository_dispatch: 'repositoryDispatch',
+    check_run: 'checkRun',
+    check_suite: 'checkSuite',
+    create: 'create',
+    delete: 'delete',
+    deployment: 'deployment',
+    deployment_status: 'deploymentStatus',
+    fork: 'fork',
+    gollum: 'gollum',
+    issue_comment: 'issueComment',
+    issues: 'issues',
+    label: 'label',
+    milestone: 'milestone',
+    page_build: 'pageBuild',
+    project: 'project',
+    project_card: 'projectCard',
+    project_column: 'projectCollumn',
+    public: 'public',
+    pull_request: 'pullRequest',
+    pull_request_review: 'pullRequestReview',
+    pull_request_review_comment: 'pullRequestReviewComment',
+    pull_request_target: 'pullRequestTarget',
+    push: 'push',
+    registry_package: 'registryPackage',
+    release: 'release',
+    status: 'status',
+    watch: 'watch',
+    workflow_run: 'workflowRun',
+  };
+
+  mapCommunityStats(githubEvent: string): string {
+    let mappedValue: string;
+    try {
+      mappedValue = this.eventMap[githubEvent];
+    } catch {
+      throw new BadRequestException('Please Provide valid Githubevent');
+    }
+    return mappedValue;
+  }
+}
--- a/src/github/communitystats-mapping.service.ts
+++ b/src/github/communitystats-mapping.service.ts
@@ -21,7 +21,7 @@ export class CommunitystatsMappingService {
     page_build: 'pageBuild',
     project: 'project',
     project_card: 'projectCard',
-    project_column: 'projectCollumn',
+    project_column: 'projectColumn',
     public: 'public',
     pull_request: 'pullRequest',
     pull_request_review: 'pullRequestReview',
--- a/src/github/communitystats-mapping.service.ts
+++ b/src/github/communitystats-mapping.service.ts
@@ -1,4 +1,5 @@
 import { Injectable, BadRequestException } from '@nestjs/common';
+import { CommunityStats } from './interfaces/github.interface';
 
 @Injectable()
 export class CommunitystatsMappingService {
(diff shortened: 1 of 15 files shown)


src/github/dto/github.dto.ts
```diff
@@ -0,0 +1,21 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+class CommunityStats {
+  [key: string]: number;
```
Cahllagerfeld, line 4: This class definition should be adjusted depending on the outcome of the question in #63

src/github/communitystats-mapping.service.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable, BadRequestException } from '@nestjs/common';
+
+@Injectable()
+export class CommunitystatsMappingService {
+  private eventMap = {
```
Cahllagerfeld, line 5: Can we hardcode the Mapping for the Events, or is there a way to do that dynamically?

eddiejaoude: I think hardcode is probably best, these are not going to change.

Cahllagerfeld:
Hi Eddie,

I introduced the Mapping discussed in #44 
Pls feel free to give some feedback when reviewing changes!

test/features/github.feature
```diff
@@ -0,0 +1,115 @@
...
+            | followers | 500                        |
+            | repos     | 32                         |
+            | event     | "push"                     |
+        Then make a DELETE request to "/github/66"
+        Then the response status code should be 404
+        And the response should contains:
+            | statusCode | 404                       |
+            | message    | "Githubprofile Not Found" |
+
+    Scenario: update githubprofile with previously used event
```
Cahllagerfeld, line 62: tests for already used, and unused Event, no duplicate

src/github/github.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    const githubProfile = this.githubProfiles.find(
+      (profile) => profile.id === id,
+    );
+    if (!githubProfile) {
+      throw new HttpException('Githubprofile Not Found', HttpStatus.NOT_FOUND);
+    }
+    return { ...githubProfile };
+  }
+
+  //TODO Event Mapping
```
eddiejaoude, line 49: Could the **todo** be written like this so that various tools pick up on them 

```ts
// @TODO: ...
```

Cahllagerfeld: to be honest I just forgot to remove it.. :sweat_smile:

src/github/dto/github.dto.ts
```diff
@@ -0,0 +1,73 @@
...
+  @ApiProperty({ required: false })
+  @IsOptional()
+  @IsNumber()
+  followers: number;
+
+  @ApiProperty({ required: false })
+  @IsOptional()
+  @IsNumber()
+  repos: number;
+  // ... any other useful info public info https://docs.github.com/en/rest/reference/users
```
Cahllagerfeld, line 67: This was a comment which was in the inital description of #44. Do we need some aditional info from Github or is the data we currently get enough?

src/github/github.service.ts
eddiejaoude: oh ok lol 😂

Cahllagerfeld:
I added Validation for the DTO of this module.
This validation probably needs to be added to the other modules as well.
That would be a new PR then I guess

src/github/dto/github.dto.ts
eddiejaoude: I had another look, I think these fields would be good too

- location
- company
- blog (I think this is their url they set)

eddiejaoude: But this could come an another PR if you prefer

Cahllagerfeld: Yes, these info would be good as well.
You want to display the location set in the Github Profile on a card? In other words do I need to geocode them as well?

Cahllagerfeld: I'll add this for now, as these shouldnt be huge extensions to the current codebase

eddiejaoude: Yep, the location will be geocoded to show on a map - but I think in this PR, only save the property `location.provided` and the geocoding data can come in another PR.

Maybe something like this...

```ts
{
   // ...
   location: {
      provided: string;
      lat: number;
      long: number;
   }
}
```

Cahllagerfeld: Okay, then I'll just save the location that is sent for now.
I started with a geocoding service when I tried that out for #38 based on observables and rxjs. (I didnt open PR yet)
We could make a Commonsmodule or sth like that where we put such "helper-services"

eddiejaoude: Yep, I don't think we need to allow them to save the `lat/lng`, this will come from the service after they provide a location string.

> I started with a geocoding service when I tried that out for #38 based on observables and rxjs. (I didnt open PR yet)

oh awesome! 👍 

> We could make a Commonsmodule or sth like that where we put such "helper-services"

I don't think this is needed for now, as this is the only place the location geocoding will be happening

Cahllagerfeld: I've got one question over: 
In #38 theres also location for the discord data.
What would happen if a user has two different location set in his profiles?
That would probably result in wrong pins on the map. 
How should we handle this? Or do we even have to handle this?

eddiejaoude: Sorry, that was my mistake 🤦‍♂️ 

We used to get it from Discord bot, but now we are planning to integrate with GitHub we don't need to get it from the Discord Bot anymore set by the user.

I create an issue for it to be removed from Discord module https://github.com/EddieHubCommunity/api/issues/66

Cahllagerfeld: Okay, that will make it a bit easier for us then :D
You see #66 as new PR, or should I put it in this?

eddiejaoude: I would prefer #66 as a new PR, the smaller the PRs the better

Cahllagerfeld: Okay, then I implement the new properties
- organization
- blog
- location (not geocoded)

and start implementing geocoding service when this is merged
Sounds like a plan!

eddiejaoude: Sounds good 👍

eddiejaoude (approved):
Looks great 👍

eddiejaoude:
Great work 💥 , thank you 🎉

PR #67, opened 2021-04-19
Issue-66 Geocodingservice

### LINKED ISSUE #66: Remove location from Discord models
Location will come from GitHub data, no need for the user to set it via the Discord Bot

Cahllagerfeld:
closes #66 
Geocode location from githubprofile to coordinates for display on a map

### DIFF
--- a/src/github/geocoding.service.ts
+++ b/src/github/geocoding.service.ts
@@ -1,4 +1,54 @@
-import { Injectable } from '@nestjs/common';
+import { HttpService, Injectable } from '@nestjs/common';
+import { map } from 'rxjs/operators';
+import { AxiosResponse } from 'axios';
+import { GithubLocation } from './interfaces/github.interface';
+import { Observable } from 'rxjs';
+
+interface Address {
+  tourism: string;
+  house_number: string;
+  road: string;
+  quarter: string;
+  city: string;
+  state_district: string;
+  state: string;
+  postcode: string;
+  country: string;
+  country_code: string;
+}
+
+interface GeocodedResponse {
+  place_id: number;
+  licence: string;
+  osm_type: string;
+  osm_id: number;
+  boundingbox: string[];
+  lat: string;
+  lon: string;
+  display_name: string;
+  class: string;
+  type: string;
+  importance: number;
+  icon: string;
+  address: Address;
+}
 
 @Injectable()
-export class GeocodingService {}
+export class GeocodingService {
+  constructor(private http: HttpService) {}
+
+  async fetchCoordinates(location: string): Promise<GithubLocation> {
+    const url = `https://nominatim.openstreetmap.org/?addressdetails=1&q=${encodeURIComponent(
+      location,
+    )}&format=json&limit=1`;
+
+    const response: AxiosResponse = await this.http.get(url).toPromise();
+    const data: GeocodedResponse = response.data[0];
+    const locationObject: GithubLocation = {
+      provided: location,
+      lat: parseFloat(data.lat),
+      long: parseFloat(data.lon),
+    };
+    return locationObject;
+  }
+}
--- a/src/github/geocoding.service.ts
+++ b/src/github/geocoding.service.ts
@@ -1,0 +1,4 @@
+import { Injectable } from '@nestjs/common';
+
+@Injectable()
+export class GeocodingService {}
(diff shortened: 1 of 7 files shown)


Cahllagerfeld:
Its currently promise-based, cause I had the feeling that Observables arent necessary at this point.
If u are in favour of using Observables I can use these, although I dont feel quite comfortable with them yet 😅

src/github/geocoding.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  quarter: string;
+  city: string;
+  state_district: string;
+  state: string;
+  postcode: string;
+  country: string;
+  country_code: string;
+}
+
+interface GeocodedResponse {
```
Cahllagerfeld, line 18: These Interfaces arent quite necessary, I just add them for consistency.
They represent the Response from the Geocoding http-call.

src/github/geocoding.service.ts
```diff
@@ -0,0 +1,52 @@
...
+@Injectable()
+export class GeocodingService {
+  constructor(private http: HttpService) {}
+
+  async fetchCoordinates(location: string): Promise<GithubLocation> {
+    const url = `https://nominatim.openstreetmap.org/?addressdetails=1&q=${encodeURIComponent(
+      location,
+    )}&format=json&limit=1`;
+
+    const response: AxiosResponse = await this.http.get(url).toPromise();
```
eddiejaoude, line 43: I think NestJS recommends to use RxJS?

Which would become something like this I think...

```ts
     return this.http.get(url)
        .pipe(
            map((response) => response.data),
            map((data) => ({
                provided: location,
                lat: parseFloat(data.lat),
                long: parseFloat(data.lon),
             })),
       )
```

eddiejaoude:
Awesome work, you are very fast 💥 . I left a comment about promises / rxjs

src/github/geocoding.service.ts
Cahllagerfeld: Hi Eddie, 

I think I need your help: 
```ts
if (body.location) {
      const locationObject = await this.geocodingService.fetchCoordinates(
        body.location,
      );
      newGithubProfile.location = locationObject;
    }

If we switch this from Promise to Observable, this if-loop needs to subscribe to geocoding-function. How is this handled in case we dont have location sent. In this case we dont have an observable. I'm not quite sure how to implement this. 😅 
```

eddiejaoude: Sure, no problem I can try to help - I am a bit out of practice with RxJS. But we will figure it out 🤓

Cahllagerfeld: Really looking forward to this.
I can change the logic to use Observables. 
This will probably break the logic as I dont know yet how to do it with rxjs.
Is this okay?

eddiejaoude: I had a brief play with the RxJS, it is not complete, but I think lets keep it as promises for now most people are not familiar with RxJS. Just to give you an example of where I got to... (the promises with async/await look simpler)

```ts
return of(newGithubProfile)
      .pipe(
        map((data) => data.username ? data : throwError(new HttpException('Incomplete Data', HttpStatus.BAD_REQUEST))),
        switchMap((data) => this.geocodingService.fetchCoordinates(body.location)
          .pipe(
            defaultIfEmpty(data),
            filter((body) => body.location),
            map((location) => {
              data.location = location;

              return data;
            })
          )
        ),
        map((data) => {
          this.githubProfiles.push(data);

          return data;
        ),
      );
```

Cahllagerfeld: Well thats probably true.
We could open an issue for switching from Promise to Observable.
I think when having a database we are going to face many other asynchronous operations..
We should validate or spike if Observables are our way to go before implementing the database otherwise we would have many things to refactor.

eddiejaoude: Sounds like a good plan 👍

Cahllagerfeld: Should I leave the coding for this PR then as it is?

eddiejaoude: Yes, I think so 👍

eddiejaoude (approved):
Great work 👍

src/github/geocoding.service.ts
eddiejaoude: I prefer being strict, so these are great 👍

PR #69, opened 2021-04-19
feat: AuthModule

### LINKED ISSUE #51: Authentication-Service
Does this Api need to be protected by JWT using `passport` for example?

### LINKED ISSUE #70: Bug: Geocoding causes Crash when not providing valid Location
The API crashes when no real Data is returned from the geocoding-Service.

I suggest adding `try-catch` adding to it.

Cahllagerfeld:
fixes #51 
fixes #70 
Authentication for Writing-Endpoints of API


Cahllagerfeld:
feature is using [this](http://www.passportjs.org/packages/passport-unique-token/) Passport-Strategy
this Strategy required an adjustment in `tsconfig.json`: The flag `esModuleInterop=true` had to be added.
Do you know how to implement e2e testing with cucumber? To implement this the Token in the Environment variables always needs to be same, so I'm not quite sure how to do that.

Cahllagerfeld:
I used the `CombinedGuard` in the current version, but I also added the single one for the Discordbot, and for the Github Action although they arent used.
Shall I delete them again until we need them, or do we keep them in case we need them in the future?
What do you think is better?

src/auth/Combined.strategy.ts
```diff
@@ -0,0 +1,25 @@
...
+) {
+  constructor(private config: ConfigService) {
+    super({});
+  }
+
+  async validate(token: string, done) {
+    if (
+      token != this.config.get('APP_DISCORD_TOKEN') &&
+      token != this.config.get('APP_GITHUB_TOKEN')
+    ) {
```
eddiejaoude, line 19: I think this check might be better as an array, rather than app specific

```ts
if (this.config.get('APPROVED_TOKENS').includes(token))
```

eddiejaoude:
Wow this is great 💥  Thank you @Cahllagerfeld

src/auth/Combined.strategy.ts
Cahllagerfeld: Okay 👍 
I did this on purpose as explained above.
If we dont have the described scenario thats probably the easier way of doing it.

eddiejaoude: Sorry I got excited and missed your explanation above 🤦‍♂️ . I think keep it simple and only what we need for now would reduce confusion for people joining the project and less for us to test.

eddiejaoude:
Btw I think our file naming convention used so far, is files start with a lowercase, but some of these files seem to start with a capital case, could they be made lowercase to be consistent?

Cahllagerfeld:
sure 👍 
I'll adjust that as well.

Cahllagerfeld:
fixes #70

Cahllagerfeld:
I dont know hat to adjust the e2e-Tests as the environmentvariables may change from instance to instacne of the app

src/github/geocoding.service.ts
```diff
@@ -42,11 +47,18 @@ export class GeocodingService {
...
-    };
-    return locationObject;
+    try {
+      const locationObject: GithubLocation = {
+        provided: location,
+        lat: parseFloat(data.lat),
+        long: parseFloat(data.lon),
+      };
+      return locationObject;
+    } catch {
```
Cahllagerfeld, line 57: This fixes #70

Cahllagerfeld:
I fixed the Unittests for now and the e2e-tests run. 
There was an issue with default-imports caused by the `esModuleInterop=true`-Flag in `tsconfig`.
I researched it and found a solution which doesnt break the default-imports: `"allowSyntheticDefaultImports": true`

I did some research about cucumber testing here are my key-takeaways: 
1. We probably need new step-defintions for authenticated requests, as we need to send a token-header
2. Setting environment variables doesnt seem to be that easy in cucumber 
image

eddiejaoude:
> I dont know hat to adjust the e2e-Tests as the environmentvariables may change from instance to instacne of the app

Sorry I did not get back to you fast enough, looks like you solved it 👇 

> I fixed the Unittests for now and the e2e-tests run.

Thank you! 👍 this is great 💥 

> Furthermore I did some research about cucumber testing, here are my key-takeaways:
> We probably need new step-defintions for authenticated requests, as we need to send a token-header
> Setting environment variables doesnt seem to be that easy in cucumber

I can take a look at this, but it might be a few days before I am able to

Cahllagerfeld:
Do not stress yourself with this.
I'll have look at this again during the day as well!

eddiejaoude:
this patch should apply cleanly to your fork and branch `issue-51` with `git apply 51.patch`, any issues let me know
PS. great work on that PR 💥 

```diff
diff --git a/package.json b/package.json
index b2b0818..32f4392 100644
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "cucumber-js -p default"
+    "test:e2e": "APPROVED_TOKENS=abc,def cucumber-js -p default"
   },
   "dependencies": {
     "@nestjs/common": "^7.6.13",
diff --git a/src/auth/discordGithub.strategy.ts b/src/auth/discordGithub.strategy.ts
index 8606843..8c18fd8 100644
--- a/src/auth/discordGithub.strategy.ts
+++ b/src/auth/discordGithub.strategy.ts
@@ -15,7 +15,7 @@ export class DiscordGithubStrategy extends PassportStrategy(
   async validate(token: string, done) {
     const approvedTokens: Array<string> = this.config
       .get('APPROVED_TOKENS')
-      .split(' ');
+      .split(',');
     if (!approvedTokens.includes(token)) {
       throw new UnauthorizedException();
     }
diff --git a/test/features/discord.feature b/test/features/discord.feature
index c9a5326..f69046c 100644
--- a/test/features/discord.feature
+++ b/test/features/discord.feature
@@ -6,7 +6,8 @@ Feature: discord module
     And the response should be "[]"
 
   Scenario: add a new user
-    Given make a POST request to "/discord" with:
+    Given authorisation
+    And make a POST request to "/discord" with:
       | bio      | "This is a GitHub Campus Expert"                                                              |
       | username | "khattakdev"                                                                                  |
       | socials  | {"discord":"khattakdev","github":"khattakdev","linkedin":"khattakdev","twitter":"khattakdev"} |
diff --git a/test/step-definitions/requests.ts b/test/step-definitions/requests.ts
index b2acb6d..ac23641 100644
--- a/test/step-definitions/requests.ts
+++ b/test/step-definitions/requests.ts
@@ -18,6 +18,11 @@ export class requests {
     await this.context.app.init();
   }
 
+  @given(/authorisation/)
+  public async authorisation() {
+    this.context.token = 'abc';
+  }
+
   @given(/make a GET request to "([^"]*)"/)
   public async getRequest(url: string) {
     this.context.response = await request(this.context.app.getHttpServer()).get(
@@ -27,9 +32,13 @@ export class requests {
 
   @given(/make a POST request to "([^"]*)" with:/)
   public async postRequestWithBody(url: string, table: { rawTable: [] }) {
-    this.context.response = await request(this.context.app.getHttpServer())
-      .post(url)
-      .send(this.context.tableToObject(table));
+    const post = request(this.context.app.getHttpServer()).post(url)
+
+    if (this.context.token) {
+      post.set('token', 'abc');
+    }
+
+    this.context.response = await post.send(this.context.tableToObject(table));
 
     this.context.preRequest = await request(
       this.context.app.getHttpServer(),
diff --git a/test/support/world.ts b/test/support/world.ts
index 708501e..94b472d 100644
--- a/test/support/world.ts
+++ b/test/support/world.ts
@@ -2,6 +2,7 @@ export default class Context {
   public app;
   public response;
   public preRequest;
+  public token;
 
   public tableToObject(table) {
     return table.rawTable.reduce((result, current) => {
```

Cahllagerfeld:
I just need to copy this to a file with `.patch` right?
I get an error when doing so.. It says `error: corrupt patch at line 85`, which basically is the last line of this patch.

eddiejaoude:
Do you have any changes locally? This could cause the patch to fail. I created it from your latest commit you pushed to your `issue-51` branch. I just tested the patch against that branch and latest commit with no error, maybe you could share your console?

Here is what I did...

Pasted diff content from above into `51.patch`
```
➜  api-cahl git:(issue-51) vim 51.patch 
```

Check there are no new changes locally and apply the patch
```
➜  api-cahl git:(issue-51) git apply 51.patch 
```

Now there should be changes locally from the patch
```
➜  api-cahl git:(issue-51) ✗ git status -s
 M package.json
 M src/auth/discordGithub.strategy.ts
 M test/features/discord.feature
 M test/step-definitions/requests.ts
 M test/support/world.ts
```

I can record my screen if that helps?

Cahllagerfeld:
I copied this to 51.patch
grafik

then I get this result 
```
$ git apply 51.patch
error: corrupt patch at line 85
```
I dont have any local changes :(

eddiejaoude:
Oh that is strange, can you take a screenshot of what you have on line `85`?

Cahllagerfeld:
Its the last line of the patch 
grafik

eddiejaoude:
Hmm it looks the same as mine 🤔 the only difference is I have a blank line at the end, I know Linux and GitHub recommend every file to have a blank line at the end with no spaces, but I never thought it would cause something break.

Screenshot 2021-04-21 at 11 53 28

Cahllagerfeld:
Thats strange haha with a blank line at the end it works :o

eddiejaoude:
lol oh that's good, I learnt something new 🤓

Cahllagerfeld:
Unfortunately I run into an issue when running `npm run test:e2e`
I would not rule out the possibility that this is a problem with the operating system. You're using Mac as far as I know, I'm having windows.
```
$ npm run test:e2e

> api@0.1.1 test:e2e C:\Users\Julian\Documents\GitHub\eddiehub-api
> APPROVED_TOKENS=abc,def cucumber-js -p default

Der Befehl "APPROVED_TOKENS" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! api@0.1.1 test:e2e: `APPROVED_TOKENS=abc,def cucumber-js -p default`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the api@0.1.1 test:e2e script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Julian\AppData\Roaming\npm-cache\_logs\2021-04-21T12_00_48_050Z-debug.log
```
Translated it means sth like `the command 'APPROVED_TOKENS' is neither writen wrong or couldnt be found`

eddiejaoude:
Oh yes, it will be the way I set the environment variable for my Mac (linux style) with `APPROVED_TOKENS=abc,def npm run test:e2e`. I don't use windows, but maybe we can use the config file from NestJS for that?

eddiejaoude:
I just tested it and we can use `.env`. I can send you a new patch file or you can make these changes...

1. `package.json:20` back to what it was before with `"test:e2e": "cucumber-js -p default"`
2. `.env` file to have `APPROVED_TOKENS=abc,def`

Cahllagerfeld:
We would have to commit the `.env` then to the repository not just the `.env.example`.
Well these arent any secret data, its just important then that you use other tokens when taking it to production

eddiejaoude:
I still think we should not commit the `.env` file, because the first step for anyone or CI (GitHub Actions) will be to duplicate the `.env.template` file and make it their own.

Cahllagerfeld:
Okay this seems legit.
I'll implement it as you said :D
We just have to make sure the tests pass at everyone who develops + the ci needs to pass as well.

eddiejaoude:
yep 👍 I think using the `.env` with NestJS should work for everyone and we will need to make a small tweak to the GitHub Action to pass on CI

.env.example
```diff
@@ -0,0 +1 @@
+APPROVED_TOKENS=<Tokens seperated by comma>
```
eddiejaoude, line 1: It is better for every file to have a blank line at the end

src/github/geocoding.service.ts
eddiejaoude: Awesome! Thanks!

eddiejaoude:
This is really coming together and looking great 👍

.env.example
Cahllagerfeld: Oh yes :D Lessons learned from this afternoon :point_up: :D

Cahllagerfeld:
If we go with this solution, I'll adjust the e2e tests then.

We should definitly document this Authentication thing for e2e-Tests.
Is there already anything planned regarding documentation of the API, or is Swagger enough?

.env.example
eddiejaoude: 🤓

eddiejaoude:
> If we go with this solution, I'll adjust the e2e tests then.

Do you feel/think something is missing or there is a better way to do it?

> We should definitly document this Authentication thing for e2e-Tests.
> Is there already anything planned regarding documentation of the API, or is Swagger enough?

I do love documentation, for the end points I think Swagger is enough, but I think we could have some more info on the authentication side with some sample requests (we can help with this too 🤓 )

Cahllagerfeld:
Im very happy with this solution as its plattform agnostic and relatively easy to implement (at least in theory) 😄

Cahllagerfeld:
>I do love documentation, for the end points I think Swagger is enough, but I think we could have some more info on the authentication side with some sample requests (we can help with this too 🤓 )
We could provide a `rest.http`-File for additional Requests.

Cahllagerfeld:
With `.env-File` I managed to get them all passed locally. 
Now we just need to adjust the actions and we should be ready for reviewing this PR.

.github/workflows/build.yml
```diff
@@ -27,6 +27,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
+      - name: Create .env file
```
Cahllagerfeld, line 30: Its just a proposal: 

But maybe this way we can fix e2e-Tests finally

eddiejaoude: I think as part of the `npm install` we can use the `postinstall` hook to copy the `.env.template` file to `.env`, that way it will work for people locally too

Cahllagerfeld: I didnt think about that. This is actually much better than using a custo action for this. 
I'll revert this and add a `postinstall hook to package.json`.

eddiejaoude: Good point 👍 I would use a npm package to do the copy to solve this issue, something like https://www.npmjs.com/package/copyfiles

postinstall.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
```
Cahllagerfeld, line 1: Theres no extra module needed in theory. With this small script it works just fine.
Im still not 100% sure if thats what we want.

Cahllagerfeld: The tests now passed, but thers an linting error. Ill fix this tomorrow. Lets review it tomorrow as well!

eddiejaoude: Not sure if another script is more confusing? What command did you use with that npm package? I think for the last param you cold use a filename or a `.`?

Cahllagerfeld: I used `copyfiles .env.example .env` 😅 I'll test it tomorrow because I already turned off my computer for today 😅

eddiejaoude: >  I'll test it tomorrow because I already turned off my computer for today

yes it is late, no rush, get some rest. Thank you

Cahllagerfeld:
FINALLY it works!

eddiejaoude:
Awesome 🎉 

Sorry to be a pain, looking at that npm library, the maintainer has archived the project and recommends using alternatives instead  https://github.com/screendriver/cp-cli#this-github-project-has-been-archived-in-favour-of

src/auth/discordGithub.strategy.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { AuthGuard, PassportStrategy } from '@nestjs/passport';
+import { UniqueTokenStrategy } from 'passport-unique-token';
+
+@Injectable()
+export class DiscordGithubStrategy extends PassportStrategy(
```
eddiejaoude, line 7: I think this name should be more generic, because it will be used in other modules also 🤔

Cahllagerfeld: We could rename this to `TokenStrategy`?
Or what would you suggest? :D

test/features/discord.feature
```diff
@@ -6,7 +6,8 @@ Feature: discord module
     And the response should be "[]"
 
   Scenario: add a new user
-    Given make a POST request to "/discord" with:
+    Given authorisation
+    And make a POST request to "/discord" with:
```
eddiejaoude, line 10: We should probably have 2 more tests:

- get list of users with authorisation token and check it still works
- add a new user without authorisation token checking it fails correctly

test/features/github.feature
```diff
@@ -6,7 +6,8 @@ Feature: github-module
         And the response should be "[]"
 
     Scenario: add a new githubprofile
-        Given make a POST request to "/github" with:
+        Given authorisation
```
eddiejaoude, line 9: Same as here https://github.com/EddieHubCommunity/api/pull/69/files#r618151438

test/features/standup.feature
```diff
@@ -6,7 +6,8 @@ Feature: Standup-module
         And the response should be "[]"
 
     Scenario: add a new standup
-        Given make a POST request to "/standup" with:
+        Given authorisation
+        And make a POST request to "/standup" with:
```
eddiejaoude, line 10: Same as here https://github.com/EddieHubCommunity/api/pull/69/files#r618151438

test/step-definitions/requests.ts
```diff
@@ -27,9 +32,13 @@ export class requests {
 
   @given(/make a POST request to "([^"]*)" with:/)
   public async postRequestWithBody(url: string, table: { rawTable: [] }) {
-    this.context.response = await request(this.context.app.getHttpServer())
-      .post(url)
-      .send(this.context.tableToObject(table));
+    const post = request(this.context.app.getHttpServer()).post(url);
+
+    if (this.context.token) {
+      post.set('token', 'abc');
```
eddiejaoude, line 38: I think this was my mistake, but this should be

```ts
post.set('token', this.context.token);
```

test/step-definitions/requests.ts
```diff
@@ -38,15 +47,25 @@ export class requests {
 
   @when(/make a PUT request to "([^"]*)" with:/)
   public async putRequest(url: string, table: { rawTable: [] }) {
-    this.context.response = await request(this.context.app.getHttpServer())
-      .put(url)
-      .send(this.context.tableToObject(table));
+    const putReq = request(this.context.app.getHttpServer()).put(url);
+
+    if (this.context.token) {
+      putReq.set('token', 'abc');
```
eddiejaoude, line 53: I think this was my mistake, but this should be

```ts
post.set('token', this.context.token);
```

test/step-definitions/requests.ts
```diff
@@ -38,15 +47,25 @@ export class requests {
...
 
   @when(/make a DELETE request to "([^"]*)"/)
   public async deleteRequest(url: string) {
-    this.context.response = await request(
-      this.context.app.getHttpServer(),
-    ).delete(url);
+    const deleteReq = request(this.context.app.getHttpServer()).delete(url);
+
+    if (this.context.token) {
+      deleteReq.set('token', 'abc');
```
eddiejaoude, line 66: I think this was my mistake, but this should be

```ts
post.set('token', this.context.token);
```

src/auth/discordGithub.strategy.ts
eddiejaoude: Yep I think you are right, `TokenStrategy` looks good 👍

eddiejaoude:
Sorry I left a few more comments, then I think we are ready to merge - this PR is amazing 💥

Cahllagerfeld:
I hope I didnt forget anything

eddiejaoude (approved):
Amazing 👍 thank you!! 🔥

Cahllagerfeld:
I cant believe it :o most intese PR I've done so far :D 
Many Many thank for your support !

PR #74, opened 2021-04-23
Logging

### LINKED ISSUE #60: Centralized Logging
When growing the Application we might need some Monitoring about Performance, Failures or Errors. We could potentially implement centralized logging to Azure App Insights or Sentry. 
If we do so, we could also implement this to the different Frontends like EddieBot or a Website

Cahllagerfeld:
fixes #60 
Uses the build-in Nest-Logger.
In case we want to extend Logging we should use a external logging-package like `winston`.

Logging can be turned on by setting environmentvariable `DEBUG=true`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 APPROVED_TOKENS=abc,def
+DEBUG=false
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ $ npm run test:e2e
 $ npm run test:cov
 ```
 
+## Logging
+
+Logging can be turned on by settings an environmentvariable `DEBUG=true`
+If this variables is not true or undefiend no Logs will be made.
+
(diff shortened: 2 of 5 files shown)


eddiejaoude:
Looks great 👍 . Could you add a screenshot with and without `DEBUG=true`

Also could info be added to the README about this feature please, because I know I will forget in the future lol 😂

Cahllagerfeld:
When Logging is disabled, there are just the Logs Nest.js makes when starting the App
image

When Logging is enabled you will see ALL incoming Requests with the scheme 
`<statusCode> | [<method> <originalUrl> - <body>`

115776499-78bc9e00-a3b4-11eb-9340-f8e887dc607a

eddiejaoude (approved):
Looks good, thank you 👍

PR #76, opened 2021-04-23
issue-58--Calendar-Module

### LINKED ISSUE #58: Idea: Calendar Module
In case of having a "Community Calendar" we could integrate this to the Api, to access Data from that Calendar in other projects as well.
Before realizing that we should definitly specify some more Details - Feel free to Add Ideas in the Comments!

### LINKED ISSUE #75: Use Helmet
We should use helmet to improve security for our api.
Its like a default-package when building an api with express.
[Helmet](https://docs.nestjs.com/security/helmet)

Cahllagerfeld:
fixes #58 
fixes #75 

Calendar Module for saving community Events

- [x] CRUD Operations are ready.
- [x] e2e-Tests still missing

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,11 +3,13 @@ import { NestFactory } from '@nestjs/core';
 import { SwaggerModule } from '@nestjs/swagger';
 import { AppModule } from './app.module';
 import { swaggerConfig } from './swagger';
+import * as helmet from 'helmet';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  app.useGlobalPipes(new ValidationPipe());
+  app.useGlobalPipes(new ValidationPipe({ transform: true }));
+  app.use(helmet());
 
   const document = SwaggerModule.createDocument(app, swaggerConfig);
   SwaggerModule.setup('swagger', app, document);
--- a/test/features/calendar.feature
+++ b/test/features/calendar.feature
@@ -1,0 +1,165 @@
+Feature: calendar module
+
+    Scenario: get list of events
+        Given make a GET request to "/calendar"
+        Then the response status code should be 200
+        And the response should contains:
+            | future  | [] |
+            | ongoing | [] |
+
+    Scenario: add a new user
+        Given authorisation
+        And make a POST request to "/calendar" with:
+            | name        | "Livestream XY"                       |
+            | description | "descriptive Description"             |
+            | url         | "https://domain.com"                  |
+            | platform    | "YouTube"                             |
+            | author      | {"platform":"discord","uid":"hubber"} |
+            | startDate   | "2021-01-01T00:00:00.000Z"            |
+            | endDate     | "2021-01-01T00:00:00.000Z"            |
+
+        Then the response status code should be 201
+        And the response should contains:
+            | id          | 123                                   |
+            | name        | "Livestream XY"                       |
+            | description | "descriptive Description"             |
+            | url         | "https://domain.com"                  |
+            | platform    | "YouTube"                             |
+            | author      | {"platform":"discord","uid":"hubber"} |
+            | startDate   | "2021-01-01T00:00:00.000Z"            |
+            | endDate     | "2021-01-01T00:00:00.000Z"            |
+            | createdOn   | "2021-01-01T00:00:00.000Z"            |
+            | updatedOn   | "2021-01-01T00:00:00.000Z"            |
+
+    Scenario: add an empty event
+        Given authorisation
+        And make a POST request to "/calendar" with:
+            | test | "test" |
+        Then the response status code should be 400
+        And the response should contains:
+            | statusCode | 400               |
+            | message    | "Incomplete Data" |
+
+
+    Scenario: update an event
+        Given authorisation
+        And make a POST request to "/calendar" with:
+            | name        | "Livestream XY"                       |
+            | description | "descriptive Description"             |
+            | url         | "https://domain.com"                  |
+            | platform    | "YouTube"                             |
+            | author      | {"platform":"discord","uid":"hubber"} |
+            | startDate   | "2021-01-01T00:00:00.000Z"            |
+            | endDate     | "2021-01-01T00:00:00.000Z"            |
+        When make a PUT request to "/calendar/123" with:
+            | name        | "Livestream YZ"                      |
+            | description | "undescriptive Description"          |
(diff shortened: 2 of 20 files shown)


Cahllagerfeld:
- [x] CRUD Operations are ready.
- [ ] e2e-Tests still missing

Cahllagerfeld:
image
I implemented the response for the `getAll`-Endpoint like shown above:
- The Events are already sorted in future and ongoing, so the client doesnt need to implement this logic.
- When using a database we should define the timeranges before making the request and not filtering like here. Otherwise we have to read **ALL** Entries for events that have ever been made and this may be a huge payload.

eddiejaoude:
Looking great 💥

Cahllagerfeld:
Thanks :D I'll finish the e2e-Tests tomorrow.
I think when we got this module, theres just the database-thing open until we can make a release with the minimal set of requirements.
Are there any other feature-requests yet for future releases?

eddiejaoude:
> Are there any other feature-requests yet for future releases?

For the calendar or generally?

Cahllagerfeld:
I thought about general.
But I realized that this database stuff might be a bigger thing, than I initially thought, So I probably should rather focus on our open issues thanon future ones 😅

Cahllagerfeld:
Dont merge this yet pls 😄 
I found a little bug regarding the validation pipe.
The tests are passing for now but I need to add something :D

Cahllagerfeld:
I added the validationpipe for every dto.
Im currently implementing this validationpipe for the e2e testing which means I have to change response body for validation-errors.

My question now is:
When we use the `put`-Endpoint. Do we want to transfer the same object as when creating the entity, or do we just want to transfer the properties that have changed?

eddiejaoude:
> When we use the put-Endpoint. Do we want to transfer the same object as when creating the entity, or do we just want to transfer the properties that have changed?

Great question! For a `put` it should be the whole object, any missing fields should be deleted from the DB. We should use `patch` to only update the sent fields and leave the other as they are

Cahllagerfeld:
Okay, then fortunately I dont need to change that much 😄 
I'll open an issue for Patch then

Cahllagerfeld:
Added the Validation Pipe.
I think this PR is ready for review then :)

src/main.ts
```diff
@@ -3,11 +3,13 @@ import { NestFactory } from '@nestjs/core';
...
 import { AppModule } from './app.module';
 import { swaggerConfig } from './swagger';
+import * as helmet from 'helmet';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  app.useGlobalPipes(new ValidationPipe());
+  app.useGlobalPipes(new ValidationPipe({ transform: true }));
+  app.use(helmet());
```
Cahllagerfeld, line 12: Helmet is added here

eddiejaoude: Awesome 👍

test/features/calendar.feature
```diff
@@ -0,0 +1,167 @@
...
+        And the response should contains:
+            | statusCode | 400                                                                                                                                                                                                                                                                                                                                              |
+            | message    | ["name must be a string","name should not be empty","platform must be one of the following values: YouTube, Twitch","platform must be a string","platform should not be empty","author should not be empty","startDate should not be empty","startDate must be a Date instance","endDate must be a Date instance","endDate should not be empty"] |
+            | error      | "Bad Request"                                                                                                                                                                                                                                                                                                                                    |
+
+
+
+    Scenario: update an event
```
eddiejaoude, line 46: Do we also need a test to update a non-existing event?

test/features/calendar.feature
```diff
@@ -0,0 +1,167 @@
...
+            | description | "undescriptive Description"          |
+            | url         | "https://mydomain.com"               |
+            | platform    | "Twitch"                             |
+            | author      | {"platform":"discord","uid":"hubby"} |
+            | startDate   | "2021-01-01T00:00:00.000Z"           |
+            | endDate     | "2022-01-01T00:00:00.000Z"           |
+        When make a GET request to "/calendar"
+        Then the response status code should be 200
+        And the response should contains:
+            | future  | [{"id":123,"name":"Livestream XY","description":"descriptive Description","url":"https://domain.com","platform":"YouTube","author":{"platform":"discord","uid":"hubber"},"startDate":"2022-01-01T00:00:00.000Z","endDate":"2023-01-01T00:00:00.000Z","createdOn":"2021-01-01T00:00:00.000Z","updatedOn":"2021-01-01T00:00:00.000Z"}]   |
```
eddiejaoude, line 166: I think we need to create another step def similar to this one that would make it easier to read - I can do it in another PR after this one is merged.

```ts
@then(/the response in item "([^"]*)" should contains:/)
public dataResponseItemTable(item: number, table: { rawTable: [] }) {
    const data = this.context.tableToObject(table);
    expect(JSON.parse(this.context.response.text)[item]).to.to.eql(data);
}
```
https://github.com/EddieHubCommunity/api/blob/main/test/step-definitions/responses.ts#L31

The new step def would be something like this

```ts
@then(/the response property "([^"]*)" in field "([^"]*)" should contains:/)
public dataResponseItemTable(property: string, item: number, table: { rawTable: [] }) {
    const data = this.context.tableToObject(table);
    expect(JSON.parse(this.context.response.text)[property][item]).to.eql(data);
}
```

The tests would be something like

```gherkin
And  the response property "future" in item "0" should contains:
            | id   | 123                |
            | name | "Livestream XY"    |
And  the response property "ongoing" in item "0" should contains:
            | id   | 123                |
            | name | "Livestream YZ"    |
```

Note: I think `contains` should be singular 🤔 - this can be changed later

eddiejaoude:
Looks great 👍 just [1 comment](https://github.com/EddieHubCommunity/api/pull/76#pullrequestreview-644109991) about 1 more test

My other comment I can do in another PR after this one, it was just for some information

Cahllagerfeld:
> Looks great +1 just [1 comment](https://github.com/EddieHubCommunity/api/pull/76#pullrequestreview-644109991) about 1 more test
> 
> My other comment I can do in another PR after this one, it was just for some information

Okay I'll add this test :smile:

eddiejaoude (approved):
Looks great, awesome work, thank you 👍

Cahllagerfeld:
You're welcome :D I enjoyed it as always 😄

PR #86, opened 2021-05-02
Issue #85 - Connect Api to Database

### LINKED ISSUE #85: Connect Module to Datastax Astra
The actual Database needs to be connected to our API

Cahllagerfeld:
fixes #85 
Connect Current API to Datastax Astra


.github/workflows/build.yml
```diff
@@ -4,31 +4,46 @@ on: [push, pull_request]
...
+      contents: read
+      packages: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
       - name: install dependencies
         run: npm ci
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
Cahllagerfeld, line 46: this seems not to work for authentication against ghpr

.npmrc
```diff
@@ -0,0 +1 @@
+@cahllagerfeld:registry=https://npm.pkg.github.com
```
Cahllagerfeld, line 1: i'm not sure if it's a good idea, to commit the `.npmrc` - Theres no token inside it but I'm still not sure.

.github/workflows/build.yml
eddiejaoude: I will do some research too

eddiejaoude:
Looks good 👍 , we will also need Stargate as a service on GitHub Actions https://docs.github.com/en/actions/guides/about-service-containers

Cahllagerfeld:
> 
> 
> Looks good 👍 , we will also need Stargate as a service on GitHub Actions https://docs.github.com/en/actions/guides/about-service-containers

Thats true. I didnt think about this yet :D I guess this might get a quite big PR..
Really looking forward to get this up and running!!

eddiejaoude:
> Really looking forward to get this up and running!!

Me too 👍  I think this PR will be a game changer for the project 💥

.npmrc
eddiejaoude: I was thinking the same, I am not sure, I think it is ok, because there is nothing private and will not effect any other project

eddiejaoude: From GitHub docs...

> You also need to add the .npmrc file to your project so all requests to install packages will go through GitHub Packages

eddiejaoude:
Btw to make the package public, is there a similar option to this? https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#configuring-visibility-of-container-images-for-your-personal-account

Cahllagerfeld:
> 
> 
> Btw to make the package public, is there a similar option to this? https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#configuring-visibility-of-container-images-for-your-personal-account

In my `Danger Zone` I only have the option for deleting the package, not making it public :(

eddiejaoude:
For my containers to make them public I had to use a "preview" feature

> Note: The Container registry is currently in public beta and subject to change. During the beta, storage and bandwidth are free. To use the Container registry, you must enable the feature preview.

Now sure if you get an option for the npm packages?

Screenshot 2021-05-02 at 10 00 51

Cahllagerfeld:
> 
> 
> For my containers to make them public I had to use a "preview" feature
> 
> > Note: The Container registry is currently in public beta and subject to change. During the beta, storage and bandwidth are free. To use the Container registry, you must enable the feature preview.
> 
> Now sure if you get an option for the npm packages?
> 
> Screenshot 2021-05-02 at 10 00 51
> 
> then with options ...
> 
> Screenshot 2021-05-02 at 10 02 39

I activated this feature as well, but it seems like Github makes a difference between containers and npm-packages :/

eddiejaoude:
> I activated this feature as well, but it seems like Github makes a difference between containers and npm-packages :/

oh 😢

.github/workflows/build.yml
```diff
@@ -4,31 +4,52 @@ on: [push, pull_request]
 jobs:
   lint:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
+      - name: Authenticate with GitHub package registry
```
Cahllagerfeld, line 13: This fixes the Installation

Cahllagerfeld:
- [ ] E2E-Tests fail, because Astra-Module cant connect to Stargate.
- [x] Unit-Tests pass

.github/workflows/build.yml
eddiejaoude: ok great 👍

.github/workflows/build.yml
```diff
@@ -4,31 +4,52 @@ on: [push, pull_request]
 jobs:
   lint:
     runs-on: ubuntu-latest
+    permissions:
```
eddiejaoude, line 7: Stargate container will need to be added, probably something like this...

```yaml
services:
      stargate:
        image: stargate
        ports:
          - 8080:8080
          - 8081:8081
          - 8082:8082
        env:
          CLUSTER_NAME: stargate
          CLUSTER_VERSION: 3.11
          DEVELOPER_MODE: true
```

eddiejaoude:
> Strange thing is they pass locally although the docker container is turned off

oh really? 😱 that is strange 🤔

.github/workflows/build.yml
```diff
@@ -4,31 +4,63 @@ on: [push, pull_request]
...
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
       - name: run build
         run: npm run build
 
   test:
+    permissions:
+      contents: read
+      packages: read
     runs-on: ubuntu-latest
+    services:
```
Cahllagerfeld, line 44: This makes them pass the e2e test

eddiejaoude: awesome, it worked 👍

eddiejaoude:
All checks are green 🎉

Cahllagerfeld:
From now on I'd suggest making small PRs for the modules as discussed in #85.
I think when we start implementing the DB we might run into the problem with the creation of the keyspace within the docker stargate instance.
Do you want to discuss this in discord, so we dont spam the comments here or doesnt it matter?

eddiejaoude:
> Do you want to discuss this in discord, so we dont spam the comments here or doesnt it matter?

I think here is good, so we can refer to it in the future and also get other opinions 

Btw is this PR still in draft?

Cahllagerfeld:
Okay thats fine :D 
I think its not anymore.

eddiejaoude (approved):
Looks great 👍

PR #90, opened 2021-05-05
feat: docker compose (#89)

### LINKED ISSUE #89: docker compose file with Stargate for local dev

eddiejaoude:
closes #89 

- running Stargate locally with docker
-

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,13 @@
+version: '3.9'
+services:
+    stargate:
+        image: 'stargateio/stargate-3_11:v1.0.0'
+        ports:
+            - '8080:8080'
+            - '8081:8081'
+            - '8082:8082'
+            - '127.0.0.1:9042:9042'
+        environment:
+            - CLUSTER_NAME=stargate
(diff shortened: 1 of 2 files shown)


Cahllagerfeld:
One Action is running for about 45 min now 😅

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.9'
+services:
+    stargate:
+        image: 'stargateio/stargate-3_11:v1.0.0'
```
Cahllagerfeld, line 4: In the [docs](https://stargate.io/docs/stargate/1.0/quickstart/quick_start-document.html) they use `stargateio/stargate-3_11:v1.0.13`
Do you think this matters?

eddiejaoude: ah good spot! 👍 looks like they have bumped the build versions, probably more important when they move minor versions

eddiejaoude:
> One Action is running for about 45 min now 😅

not sure why that is 🤔

Cahllagerfeld (approved):

Issue #29, opened 2021-03-05
Husky pre commit check


eddiejaoude:
Make sure to use v5

- Lint commit messages with conventional commits http://conventionalcommits.org
- Lint code check 

https://www.npmjs.com/package/husky

bagladivyang03:
@eddiejaoude I would like to contribute to this issue, Could u pls tell me what exactly has to be done for this issue.

eddiejaoude:
@bagladivyang03 great 👍 

1. Install husky using "automatic installer" from their docs https://typicode.github.io/husky/#/?id=automatic-recommended
2. Add 2 hooks https://typicode.github.io/husky/#/?id=add-a-hook
   i. run linter
   ii. run tests
   iii.  run commit linter https://github.com/conventional-changelog/commitlint#getting-started

bagladivyang03:
@eddiejaoude So basically I have to clone this repo and do all this in that directory?

eddiejaoude:
You will need to 

- fork this repo on GitHub
- clone your fork locally
- create a branch locally

yep, then do the steps above in that directory 👍

bagladivyang03:
@eddiejaoude Thanks :)

bagladivyang03:
> @bagladivyang03 great 👍
> 
> 1. Install husky using "automatic installer" from their docs https://typicode.github.io/husky/#/?id=automatic-recommended
> 2. Add 2 hooks https://typicode.github.io/husky/#/?id=add-a-hook
>    i. run linter
>    ii. run tests
>    iii.  run commit linter https://github.com/conventional-changelog/commitlint#getting-started

Actually, I have installed husky using "automatic installer" after that I m getting confused about how to do that run linter and run test part i.e adding hooks
Could u pls help @eddiejaoude

bagladivyang03:
@eddiejaoude I made a PR could u pls check that for this issue could u pls check :)

Cahllagerfeld:
Theres a PR for this. Is this PR still relevant or do we need to do this again?

eddiejaoude:
I think best probably to do it again 👍 we need this before people start to contribute

Cahllagerfeld:
closed with #91

PR #91, opened 2021-05-06
feat: add husky (#29)

Cahllagerfeld:
closes (#29)


Cahllagerfeld:
I'm not sure if our `lint`-command does what he is intended to do: 
image

After  doing the commit my Source-Control-Tab looks like this. 

I think thats because of the `--fix`-Flag in the `lint` script.
Is this the intended behaviour?

Cahllagerfeld:
The tests fail because, it wants to install version `0.0.1` as I bumped it `0.0.2`. So `0.0.1` isnt available anymore. Tbh I wasnt aware of this problem.. 
Shall I adjust this here as well (like in #88) or what shall I do?

nhcarrigan-test:
Hmm, that suggests that the pre-commit hook isn't behaving properly.

Additionally, the `lint` script using the `--fix` flag effectively makes it useless in CI, as the Action will "fix" the files but they won't be committed.

My recommendation is to use a separate script for `--fix` and remove it from the `lint` that the CI uses.

eddiejaoude:
merged #93 👍

Cahllagerfeld:
> 
> 
> merged #93 👍

Thx!

Cahllagerfeld:
This PR is ready for review I think

package.json
```diff
@@ -12,8 +12,9 @@
     "start:dev": "nest start --watch",
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/main",
-    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\"",
```
Cahllagerfeld, line 15: I removed the `--fix` Flag here as we are already having a `format`-Script

nhcarrigan: The format script runs prettier, though, which (while also helpful) won't necessarily fix `eslint` errors AFAIK.

Cahllagerfeld: I'd suggest a script for `lint:fix` or something similar. What do you think about that?

nhcarrigan: 100% agree.

Cahllagerfeld:
After running the script I get following console: 
image
Is this a problem, or is it fine for us, that we shouldnt open issues regarding this?

nhcarrigan:
Should be fine, I think - we could potentially look at rolling TS back to 4.1.5 if issues arise.

Cahllagerfeld:
> 
> 
> Should be fine, I think - we could potentially look at rolling TS back to 4.1.5 if issues arise.

Okay, than I think the PR is reviewable 😄

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint
```
eddiejaoude, line 4: Could we run the tests as well here?

```suggestion
npm run lint
npm test
npm test:e2e
```

Cahllagerfeld: Yes, we can do this. But then we probably cant push any WIP changes (like step-definitions) for example to the PR. They have to be resolved locally then, or am I wrong?

nhcarrigan-test: Ordinarily yes, though you can pass the `--no-verify` flag to skip the pre-commit hooks.

nhcarrigan: Ordinarily yes, though you can pass the --no-verify flag to skip the pre-commit hooks.

Cahllagerfeld: Added the scripts to the hook

PR #105, opened 2021-05-15
feat: calendar-module to database

### LINKED ISSUE #97: Connect calendar-module to database

Cahllagerfeld:
closes #97 

- [x] actual functionality
- [x] unit tests
- [x] e2e tests

### DIFF
--- a/src/calendar/calendar.service.ts
+++ b/src/calendar/calendar.service.ts
@@ -1,11 +1,7 @@
 import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
 import { CalendarEventDTO } from './dto/calendar.dto';
-import {
-  CalendarEvent,
-  SortedEventResponse,
-} from './interfaces/calendar.interface';
-import { isFuture } from 'date-fns';
-import { concatMap, filter, map, tap } from 'rxjs/operators';
+import { CalendarEvent } from './interfaces/calendar.interface';
+import { concatMap, filter } from 'rxjs/operators';
 import {
   AstraService,
   deleteItem,
@@ -16,7 +12,6 @@ import { forkJoin, Observable } from 'rxjs';
 @Injectable()
 export class CalendarService {
   constructor(private readonly astraService: AstraService) {}
-  private calendarEvents: CalendarEvent[] = [];
 
   createCalendarEvent(
     calendarEventBody: CalendarEventDTO,
--- a/src/calendar/calendar.service.ts
+++ b/src/calendar/calendar.service.ts
@@ -5,14 +5,23 @@ import {
   SortedEventResponse,
 } from './interfaces/calendar.interface';
 import { isFuture } from 'date-fns';
+import { concatMap, filter, map, tap } from 'rxjs/operators';
+import {
+  AstraService,
+  deleteItem,
+  documentId,
+} from '@cahllagerfeld/nestjs-astra';
+import { forkJoin, Observable } from 'rxjs';
 
 @Injectable()
 export class CalendarService {
+  constructor(private readonly astraService: AstraService) {}
   private calendarEvents: CalendarEvent[] = [];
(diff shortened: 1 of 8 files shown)


test/features/calendar.feature
```diff
@@ -1,217 +1,208 @@
...
+            | message    | "Unauthorized" |
+
+# Scenario: get sorted ongoing and future events
+#     Given authorisation
+#     And make a POST request to "/calendar" with:
+#         | name        | "Livestream XY"                       |
+#         | description | "descriptive Description"             |
+#         | url         | "https://domain.com"                  |
+#         | platform    | "YouTube"                             |
+#         | author      | {"platform":"discord","uid":"hubber"} |
```
Cahllagerfeld, line 174: still commented out as we would `property` in ``property` where `field` `equals` should contain:

Thats a quite complex step-def which only occurs once

Cahllagerfeld:
I added the new step-def for nested properties.
I think it can be reviewed

src/calendar/calendar.service.ts
```diff
@@ -1,67 +1,84 @@
...
-        return event;
-      }
-      return;
+    const ongoing = this.astraService.find<CalendarEvent>({
+      startDate: { $lt: new Date() },
+      endDate: { $gt: new Date() },
     });
 
-    return { future: [...future], ongoing: [...ongoing] };
+    return forkJoin({ future, ongoing });
```
eddiejaoude, line 55: Nice 👍

test/features/calendar.feature
```diff
@@ -1,217 +1,206 @@
...
+        When make a POST request to "/calendar" with:
+            | name        | "Livestream XY"                       |
+            | description | "descriptive Description"             |
+            | url         | "https://domain.com"                  |
+            | platform    | "YouTube"                             |
+            | author      | {"platform":"discord","uid":"hubber"} |
+            | startDate   | "2021-01-01T00:00:00.000Z"            |
+            | endDate     | "2021-01-01T00:00:00.000Z"            |
+        And the response should contain:
+            | documentId | "TYPE:ID" |
```
eddiejaoude, line 16: Might be good to check the other field values

test/features/calendar.feature
```diff
@@ -1,217 +1,206 @@
...
```
eddiejaoude, line 90: I think the `POST` is not needed here, just the `PUT` to an invalid resource

test/features/calendar.feature
```diff
@@ -1,217 +1,206 @@
...
+        When make a DELETE request to "/calendar/{id}"
+        Then the response status code should be 204
```
eddiejaoude, line 117: I think we should protect the details to be the author of the resource, what do you think? this can come in another issue + PR

eddiejaoude:
Looks great 👍 some small comments

test/features/calendar.feature
Cahllagerfeld: Yes thats true. Thats actually something I didnt really think about so far

Cahllagerfeld: We would probably have to handle this with a query-parameter if this is allowed
>
>A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.
>
source: [here](https://datatracker.ietf.org/doc/html/rfc7231#page-29)

eddiejaoude: Good idea 👍 

Or maybe we only allow `admin` users to delete as that might be easier. We should probably raise an issue and add this discussion there

eddiejaoude: We can continue the discussion here https://github.com/EddieHubCommunity/api/issues/107

eddiejaoude (approved):
Looks great 👍

PR #110, opened 2021-05-18
fix: prepare-script

Cahllagerfeld:
fixes error in package.json

See [docs](https://www.npmjs.com/package/husky) for prepare-script

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\"",
     "lint:fix": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
     "test": "jest",
-    "prepare": "if test \"$HUSKY\" != \"0\" ; then husky install ; fi",
+    "prepare": "husky install",
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",


package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\"",
     "lint:fix": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
     "test": "jest",
-    "prepare": "if test \"$HUSKY\" != \"0\" ; then husky install ; fi",
+    "prepare": "husky install",
```
eddiejaoude, line 19: The extra check is needed for the Dockerfile, we do not want husky to run when building the container - unfortunately husky needs to be installed to do the check itself

Cahllagerfeld: Oh I didnt know that. When I ran the `npm install` a few minutes ago, it threw an error in that script 😅

eddiejaoude: oh really 😱 let me test it locally, did I fix 1 thing and break another 🤔

eddiejaoude: I just tested it, and it worked for me locally. What was your error?

Cahllagerfeld:
```bash
> if test "$HUSKY" != "0" ; then husky install ; fi 

""$HUSKY"" kann syntaktisch an dieser Stelle nicht verarbeitet werden. //Translation see below
npm ERR! code ELIFECYCLE
npm ERR! errno 1        
npm ERR! api@0.4.0 prepare: `if test "$HUSKY" != "0" ; then husky install ; fi`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the api@0.4.0 prepare script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Julian\AppData\Roaming\npm-cache\_logs\2021-05-18T18_30_23_252Z-debug.log
```
The marked spot is translated something like this 
`cannot be processed syntactically at this point`

eddiejaoude: Oh 🤔 . What OS are you on?

Cahllagerfeld: Win10 :D

eddiejaoude: oh ok. I am on Mac, and it works on Linux. I think there was an npm library that could do the conditional check for the environment variable

eddiejaoude: I could not find what I was looking forward. Another way might be to disable this for the Dockerfile with 

```
npm install --ignore-scripts
```

eddiejaoude: I will test this out now and share what happens 🤓

eddiejaoude:
I am going to close this in favour of #115

PR #112, opened 2021-05-20
Control who can delete or change data (#107)

### LINKED ISSUE #107: Who can update/delete data?
> I think we should protect the details to be the author of the resource, what do you think? this can come in another issue + PR

> We would probably have to handle this with a query-parameter if this is allowed

> A payload within a DELETE request message has no defined semantics;
sending a payload body on a DELETE request might cause some existing
implementations to reject the request.

> Good idea 👍

> Or maybe we only allow admin users to delete as that might be easier. We should probably raise an issue and add this discussion there

Initial conversation https://github.com/EddieHubCommunity/api/pull/105

/cc @Cahllagerfeld

Cahllagerfeld:
closes #107


Cahllagerfeld:
This PR got bigger than I thought.

I hope I didn't forget any requirements, but as far as I implemented it, it should be fine 😄
Feel free to review it please.

eddiejaoude:
Wow 💥 great work though 💪 thank you! I will review tomorrow 🤓

src/auth/token.strategy.ts
```diff
@@ -9,7 +9,7 @@ export class TokenStrategy extends PassportStrategy(
   'discordGithub-strategy',
 ) {
   constructor(private config: ConfigService) {
-    super({});
+    super({ tokenHeader: 'Client-Token' });
```
eddiejaoude, line 12: Could you explain a bit more about this and where it comes from?

eddiejaoude: I was thinking the client would send in 

```
@ApiHeader({ name: 'User-Uid', required: true })
@ApiHeader({ name: 'Platform', required: true })
```

src/calendar/calendar.service.ts
```diff
@@ -78,6 +87,19 @@ export class CalendarService {
...
 
+    if (
+      !this.validationService.validateAuthor(
+        oldDocument.author,
+        authorObject.uid,
+        authorObject.platform,
+      )
+    ) {
+      throw new HttpException(
+        `update failed: author doesn't match`,
```
eddiejaoude, line 98: would this be better as standard `'`?

src/commons/validation.service.ts
```diff
@@ -0,0 +1,18 @@
...
+    objectAuthor: CalendarEventAuthor | DiscordProfileAuthor | StandupAuthor,
+    headerUid: string | string[],
+    headerPlatform: string | string[],
+  ) {
+    if (objectAuthor.uid !== headerUid) return false;
+    if (objectAuthor.platform !== headerPlatform) return false;
+
+    return true;
+  }
+}
```
eddiejaoude, line 18: Would this be better in the `auth` module?

src/util/getAuthorFromHeaders.decorator.ts
```diff
@@ -0,0 +1,17 @@
...
+    const platform = request.headers['platform'] || undefined;
+
+    return { uid, platform };
+  },
+);
+
+export interface Author {
+  uid: string | string[] | undefined;
+  platform: string | string[] | undefined;
+}
```
eddiejaoude, line 17: What do you think about this going in the `auth` module?

test/features/calendar.feature
```diff
@@ -115,6 +117,8 @@ Feature: calendar module
             | endDate     | "2021-01-01T00:00:00.000Z"            |
         And the response should contain:
             | documentId | "TYPE:ID" |
+        Then set header "User-Uid" with value "hubber"
+        Then set header "Platform" with value "discord"
```
eddiejaoude, line 121: We probably need another test to check when the headers do not match that there is a failure and no update occurs

src/auth/token.strategy.ts
Cahllagerfeld: Thats the configuration of the [Passport-Strategy](http://www.passportjs.org/packages/passport-unique-token/) for validating the actual api-key. The header values for identifying the author of the object are validated after the token validation was successfull

src/calendar/calendar.service.ts
Cahllagerfeld: oh, yes 😅 Dont know why I used these ones..

src/commons/validation.service.ts
Cahllagerfeld: Probably yes, I did this in case we need to validate other things in the future as well

Cahllagerfeld:
I'll add the tests tomorrow.. Unfortunately had a tough day today :(

src/auth/token.strategy.ts
eddiejaoude: ok makes sense, thank you 👍

src/calendar/calendar.service.ts
eddiejaoude: lol no worries 😄

src/standup/standup.service.ts
```diff
@@ -62,6 +67,18 @@ export class StandupService {
...
         }
+        if (
+          !this.validationService.validateAuthor(
+            data.author,
+            authorObject.uid,
+            authorObject.platform,
+          )
+        ) {
+          throw new HttpException(
+            "deletion failed: author doesn't match",
```
eddiejaoude, line 78: I think we were are single quotes in the project? For example https://github.com/EddieHubCommunity/api/pull/112/files#diff-1cb9cb1e7db3765afcf2fe8e30efab22b562c878d991ad2b966dfe07447cf29bR95

Cahllagerfeld: Single quotes wont work at this point because of the `doesn't` wheres the single quote

test/features/discord.feature
```diff
@@ -37,41 +37,75 @@ Feature: discord module
...
+            | bio     | "This is a GitHub Campus Expert"                                                              |
+            | author  | {"platform":"discord","uid":"hubber"}                                                         |
+            | socials | {"discord":"khattakdev","github":"khattakdev","linkedin":"khattakdev","twitter":"khattakdev"} |
         Then the response should contain:
             | documentId | "TYPE:ID" |
+        Then set header "User-Uid" with value "hubber"
         When make a DELETE request to "/discord/{id}"
         Then the response status code should be 204
 
+    Scenario: delete a user with wrong author
```
Cahllagerfeld, line 94: added the new tests

src/standup/standup.service.ts
eddiejaoude: We could escape it, if that helps with the consistency

```suggestion
            'deletion failed: author doesn\'t match',
```

src/auth/header-validation.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@nestjs/common';
+import { CalendarEventAuthor } from '../calendar/interfaces/calendar.interface';
+import { DiscordProfileAuthor } from '../discord/interfaces/discord.interface';
+import { StandupAuthor } from '../standup/interfaces/standup.interface';
+
+@Injectable()
+export class ValidationService {
+  public validateAuthor(
+    objectAuthor: CalendarEventAuthor | DiscordProfileAuthor | StandupAuthor,
```
eddiejaoude, line 9: Rather in importing other modules to here, would it be cleaner to have an interface in this module, that the other modules inherit - that way this module doesn't need them, but the other modules will need the auth module, so it is one way rather than 2 ways (I hope that makes sense 😂 )

eddiejaoude:
I left some small comments 🤓 , but looks great 👍 - awesome work 💥

src/standup/standup.service.ts
Cahllagerfeld: Just tested it. The current prettier-config we use turns it to the current implementation when escaping it.

eddiejaoude: Ah ok 👍 thanks for testing

src/auth/header-validation.service.ts
```diff
@@ -1,12 +1,10 @@
...
-import { CalendarEventAuthor } from '../calendar/interfaces/calendar.interface';
-import { DiscordProfileAuthor } from '../discord/interfaces/discord.interface';
-import { StandupAuthor } from '../standup/interfaces/standup.interface';
+import { Author } from '../util/getAuthorFromHeaders.decorator';
 
 @Injectable()
 export class ValidationService {
   public validateAuthor(
-    objectAuthor: CalendarEventAuthor | DiscordProfileAuthor | StandupAuthor,
+    objectAuthor: Author,
```
eddiejaoude, line 7: Great 👍 Would `auth` location not be better than `util`?

Cahllagerfeld: Sure, I probably forgot this file when refactoring

src/calendar/interfaces/calendar.interface.ts
```diff
@@ -1,3 +1,5 @@
+import { Author } from '../../auth/getAuthorFromHeaders.decorator';
```
eddiejaoude, line 1: For a future idea / PR, what do you think about using path alias, something like

```
import { Author } from '@auth/getAuthorFromHeaders.decorator';
```

https://docs.nestjs.com/cli/libraries

eddiejaoude (approved):
Great work 💪

PR #115, opened 2021-05-27
fix: husky check (#114)

### LINKED ISSUE #114: Husky issue in windows
Npm pre install script has a check if husky should run by using an environment variable, but does not work on windows.

This was added as husky should not run in the Dockerfile, but the native check was not working.

Try using `--ignore-scripts` instead

eddiejaoude:
closes #114 

Another solution to #110

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,6 +2,8 @@ name: Publish Docker image and deploy
 on:
   release:
     types: [published]
+  push:
+    branches: [ issue-114 ]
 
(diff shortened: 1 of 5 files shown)


eddiejaoude:
Looks like this fix worked 👍 - the docker build still fails on the npm package in the container registry but the `husky` issue on `npm install` in the Dockerfiles is now fixed and this should work on Windows with no issue now.

Cahllagerfeld (approved):
Looks good to me 👍

PR #116, opened 2021-05-27
docker-file for private packages (#50)

### LINKED ISSUE #50: Deploy to EddieHub Kubernetes cluster using GitHub Actions
Example in http://github.com/eddiejaoude/stargate

Cahllagerfeld:
closes #50

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -84,10 +84,8 @@ jobs:
         with:
           context: .
           push: true
-          # secrets: |
-          #   "github_token=${{ secrets.GITHUB_TOKEN }}"
-          env:
-            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          secrets: |
+            "github_token=${{ secrets.GITHUB_TOKEN }}"
           tags: |
             ghcr.io/eddiehubcommunity/api:v${{ steps.package-version.outputs.current-version}}
             ghcr.io/eddiehubcommunity/api:latest
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,6 +2,8 @@ name: Publish Docker image and deploy
 on:
   release:
     types: [published]
+  push:
+    branches: [issue-50]
 
 jobs:
   build:
@@ -84,6 +86,8 @@ jobs:
           push: true
           secrets: |
             "github_token=${{ secrets.GITHUB_TOKEN }}"
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           tags: |
             ghcr.io/eddiehubcommunity/api:v${{ steps.package-version.outputs.current-version}}
             ghcr.io/eddiehubcommunity/api:latest
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -78,7 +78,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
-          password: ${{ secrets.CR_PAT }}
+          password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


Cahllagerfeld:
In my local Repo I made it run [see here](https://github.com/Cahllagerfeld/api/runs/2683693497?check_suite_focus=true). It seems like the action runs several installs in which then the token gets lost. Is this intended or correct? 😅

Cahllagerfeld:
In my local Repo I made it run [see here](https://github.com/Cahllagerfeld/api/runs/2683693497?check_suite_focus=true). It seems like the action runs several installs. Is this intended or correct? 😅 
I managed to get the install and build passed. The error I get no seems to be releated to github, not our dockerfile
image

eddiejaoude:
> It seems like the action runs several installs

Yep it runs twice, one for `development` and one for `production` - this was recommended, I can't remember why 🤦‍♂️ 

I think the `403` is from when it does the `npm install`

```
Error: buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c npm install]: exit code: 1
```

eddiejaoude:
I think I noticed my mistake, it is failing on the first `npm install` and the token is not being setup until the afterwards 🤦‍♂️

Dockerfile
```diff
@@ -1,35 +1,19 @@
...
-COPY package*.json ./
-
-
-RUN npm install --production --ignore-scripts
-
-COPY . .
-
-COPY --from=development /usr/src/app/dist ./dist
-
+COPY --from=building /usr/src/app ./
```
Cahllagerfeld, line 17: I think we dont need the whole folder, just the `node_modules` and the package.json

eddiejaoude: I think only the `dist` folder is needed https://github.com/EddieHubCommunity/api/pull/116/files/4bcdb9beeeee18eb50b42febc521384dde68f81a#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L33

Cahllagerfeld: As far as I know the node_modules arent bundled in the `dist`, but let me have a check

Cahllagerfeld: But you are right we definitely need `dist` I forgot about that folder

Cahllagerfeld: image
Just deleted `node_modules` and run the `start:prod` script.

eddiejaoude: For production build, I think everything we need should be in `dist`.

eddiejaoude: Let me try too

eddiejaoude: Oh you are right, it is needed 👍 (I was thinking of Angular 🤦‍♂️ )

.github/workflows/docker.yml
```diff
@@ -2,6 +2,8 @@ name: Publish Docker image and deploy
 on:
   release:
     types: [published]
+  push:
+    branches: [issue-50]
```
Cahllagerfeld, line 6: need to remove the trigger before merging

eddiejaoude: Did this run, because it is a in a fork I think it will not run. Should I merge it to a branch to test it properly?

Cahllagerfeld: I think that would be the best idea 👍

Dockerfile
```diff
@@ -1,35 +1,21 @@
...
-
-
-RUN npm install --production --ignore-scripts
-
-COPY . .
-
-COPY --from=development /usr/src/app/dist ./dist
-
+COPY --from=building /usr/src/app/dist ./dist
+COPY --from=building /usr/src/app/node_modules ./node_modules
```
eddiejaoude, line 18: I think these will all have dev packages, we should build for prod only

PR #118, opened 2021-06-03
fix: docker-file for private packages (#50)

eddiejaoude:


Dockerfile
```diff
@@ -1,35 +1,21 @@
...
-
-RUN npm install --production --ignore-scripts
-
-COPY . .
-
-COPY --from=development /usr/src/app/dist ./dist
-
+COPY --from=building /usr/src/app/dist ./dist
+COPY --from=building /usr/src/app/node_modules ./node_modules
+COPY --from=building /usr/src/app/package.json ./package.json
```
eddiejaoude, line 19: @Cahllagerfeld it worked 👍 , but I am not sure about this, can we keep this how it was?

Cahllagerfeld: First of all Im glad it works now :D

Do you want to revert it to this version?

```Dockerfile
FROM node:15 As development
LABEL org.opencontainers.image.source https://github.com/eddiehubcommunity/api

WORKDIR /usr/src/app

COPY package*.json ./

RUN npm install

COPY . .

RUN npm run build

FROM node:15 as production

ARG NODE_ENV=production
ENV NODE_ENV=${NODE_ENV}
ENV VERSION="v0.0.0"

RUN --mount=type=secret,id=github_token \
  cat /run/secrets/github_token
RUN echo "//npm.pkg.github.com/:_authToken=$github_token" > ~/.npmrc

WORKDIR /usr/src/app

COPY package*.json ./


RUN npm install --production --ignore-scripts

COPY . .

COPY --from=development /usr/src/app/dist ./dist

CMD ["npm", "run", "start:prod"]

```

eddiejaoude: I don't think we can revert all of it, because we need the way you handled the GitHub secret for the `npm install`, but if we can use the way the installation of npm was.

I can try something if you like?

Cahllagerfeld: sure :D

.github/workflows/docker.yml
```diff
@@ -2,6 +2,8 @@ name: Publish Docker image and deploy
 on:
   release:
     types: [published]
+  push:
+    branches: [issue-50]
```
eddiejaoude, line 6: Remove before merging

.github/workflows/docker.yml
```diff
@@ -82,8 +87,9 @@ jobs:
         with:
           context: .
           push: true
-          secrets: |
-            "github_token=${{ secrets.GITHUB_TOKEN }}"
+          build-args: 'github_token=${{ secrets.GITHUB_TOKEN }}'
+          # secrets: |
+          #   "github_token=${{ secrets.GITHUB_TOKEN }}"
```
eddiejaoude, line 92: we should remove this unused code

eddiejaoude:
@Cahllagerfeld what do you think?

Cahllagerfeld:
I think we are quite close.
The question is how can I adjust the action as its not a PR linked to my fork anymore? 😅

eddiejaoude:
I think you have write access to this repo, so you can edit this PR, let me double check

eddiejaoude:
@Cahllagerfeld you already have write access to this repo, so you can edit this branch/PR directly

Cahllagerfeld:
> 
> 
> @Cahllagerfeld you already have write access to this repo, so you can edit this branch/PR directly

You're right I forgot about that...
I made the adjustments

eddiejaoude:
Great thank you @Cahllagerfeld 👍

PR #124, opened 2021-06-06
fix: working on the GitHub connection of a discord user (#108)

Souvikns:


Souvikns:
@eddiejaoude one question, All we need is the user id to fetch the connection list. Will just create the route to fetch the GitHub account link

eddiejaoude:
I think we should save it in the DB, to keep it more performant and also not to hit any rate limits on GitHub's API.

We can think of another trigger later on to keep the data up to date, but for now, I suggest if there is no GitHub data in the DB then fetch it.

What do you think @Cahllagerfeld ?

Souvikns:
Sorry for the late reply I took my vaccine and wasn't feeling well. 

So would it be better when in the API the create function in the `discord.service.ts` file is called in the POST method we make a call to the discord API and update all the socials. 

https://github.com/EddieHubCommunity/api/blob/eb559302ebb3da5d7344a9309b0c18d7e769bbbd/src/discord/discord.service.ts#L22

Cahllagerfeld:
I think we could create an endpoint in the discord-controller which is called when somebody types `!github` to the discord bot.
If the connection exists I'd update (or create) the user who called the discord command in the db as a `discord-user`. 
I think when doing this we need to update the `discord-user-interface`. The Discord DTO should be fine as the `!github` command should be the only possibility to "connect" your discord with the github id.

eddiejaoude:
> Sorry for the late reply I took my vaccine and wasn't feeling well.

no problem, I hope you are feeling better @Souvikns 

> I think we could create an endpoint in the discord-controller which is called when somebody types !github to the discord bot.

Great idea! 👍

Souvikns:
Can you give an insight into the `@AuthorObject` and what it is?

Cahllagerfeld:
[AuthorObject](https://github.com/EddieHubCommunity/api/blob/eb559302ebb3da5d7344a9309b0c18d7e769bbbd/src/auth/getAuthorFromHeaders.decorator.ts#L4) is a decorator which extracts header values from the request before passing them to the controller or service.

package.json
```diff
@@ -33,6 +33,7 @@
     "@nestjs/passport": "^7.1.5",
     "@nestjs/platform-express": "^7.6.13",
     "@nestjs/swagger": "^4.8.0",
+    "axios": "^0.21.1",
```
Cahllagerfeld, line 36: I dont think we need axios as a new dependency here. We can use the `Nest-Http-Module` like [here](https://github.com/EddieHubCommunity/api/blob/eb559302ebb3da5d7344a9309b0c18d7e769bbbd/src/github/geocoding.service.ts#L48)

Souvikns: oh ok, I am kinda new to nestjs. will change this.

Souvikns:
I was checking the discord API, hitting it with postman using my discord id but the thing is I am getting `401: unauthorized`, I think I need to pass some API key or use discord OAuth. 

Also I was checking `http://api.eddiehub.org/discord` and was getting `{"statusCode":500,"message":"Internal server error"}`

Cahllagerfeld:
[Here](https://github.com/EddieHubCommunity/api/issues/108#issue-894244356) Eddie mentioned that we need a Token which is authorized to access the connections. I'm not quite sure how to retrieve this token to be honest though.

Souvikns:
I was trying to fetch the connection data. First I tried using `discord-js` but I think it's not possible. After searching a lot I think the only way is to use discord OAuth but for some reason, it is not working for me. @Cahllagerfeld can you check it once 

I found some of links 
https://stackoverflow.com/questions/54045938/how-to-get-user-id-from-oauth-on-discord

Cahllagerfeld:
I just read the [official Discord Auth Documentation](https://discord.com/developers/docs/topics/oauth2). It seems like we need to authenticate the user to get an accesstoken we can use to get the connections he has in his profile on his behalf. I'm not quite sure how we need to handle this as we only provide the api which should use the token and not retrieve it. What do you think @eddiejaoude ?

eddiejaoude:
Sorry to reply so slowly. I will try to address each point, if I miss anything please let me know.

> I was checking the discord API, hitting it with postman using my discord id but the thing is I am getting 401: unauthorized, I think I need to pass some API key or use discord OAuth.

I think if we get the GitHub username with my idea below, this might no longer be an issue? Because we will have the string to go to GitHub to fetch the data. Hopefully I understand the situation correctly.

> I was trying to fetch the connection data. First I tried using discord-js but I think it's not possible. After searching a lot I think the only way is to use discord OAuth but for some reason, it is not working for me

>  It seems like we need to authenticate the user to get an accesstoken we can use to get the connections the user has in the profile on their behalf

I think we keep it simple and use the username they provided. This will be stored in the database here https://github.com/EddieHubCommunity/api/blob/main/src/discord/interfaces/discord.interface.ts#L15

PR #135, opened 2021-06-24
fix: no data, empty result (#134)

### LINKED ISSUE #134: No data returns a 500
I think the `500` is because there is no data in the DB and we do not handle the `404` exception. We will need something like this...

```ts
findAll(): Observable<findResult<DiscordProfile>> {
      return this.astraService.find<DiscordProfile>()
        .pipe(
          catchError((e) => from([]))
        );
  }
```

eddiejaoude:
closes #134


Cahllagerfeld (approved):
I'm really no pro with `rxjs` to be honest, but I thimk the changes look good to me

eddiejaoude:
Thank you for reviewing @Cahllagerfeld

PR #142, opened 2021-06-26
set collection and namespace dynamically (#137)

### LINKED ISSUE #137: Using namespace for separate of data
I was thinking, the `namespace` could be set in the service? this could be the name of the Discord server id - then it would help us easily separate the data between servers

Currently it is set in the module

```ts
AstraModule.forFeature({ namespace: 'eddiehub', collection: 'discord' }),
```

What do you think?

Cahllagerfeld:
closes #137


Cahllagerfeld:
I adjusted the Astra-Service from the package and updated the standup-Service with it. For me locally this worked perfectly fine. Maybe thats our solution

Cahllagerfeld:
Theres still some hardcoded stuff in it (all in all its still a bit dirty) but for me its like a proof-of-concept. If we decide to do it this way, we can pair on that if you want to!

eddiejaoude:
I think it looks great 👍 , definitely has the flexibility we need 💥

src/astra/astra.service.ts
```diff
@@ -0,0 +1,147 @@
...
+} from '@cahllagerfeld/nestjs-astra';
+const DATASTAX_CLIENT = 'DATASTAX_CLIENT';
+
+@Injectable()
+export class AstraService {
+  private collection: any;
+  constructor(@Inject(DATASTAX_CLIENT) private readonly client: any) {}
+
+  private setupClient(namespace: string, collection: string) {
+    this.collection = this.client.namespace(namespace).collection(collection);
```
eddiejaoude, line 16: I think we could try to save these into the collection with a unique key, but not sure if that helps with efficiency because it is a rest call.

Something like ...

```ts
private getClient(namespace: string, collection: string) {
   const key = `${namespace}-${collection}`;
   if (!this.collection[key]) {
      this.collection[key] = this.client.namespace(namespace).collection(collection);
   }

   return this.collection[key];
}
```

Cahllagerfeld:
The tests will fail, because I didnt adjust the e2e-tests yet. But with this commit the standup-service gets configured by a header

eddiejaoude:
Sure, awesome 👍

src/standup/standup.controller.ts
```diff
@@ -23,24 +24,31 @@ export class StandupController {
...
   @UseGuards(TokenGuard)
   @ApiSecurity('token')
-  createStandup(@Body() body: StandupDTO) {
-    return this.standupService.create(body);
+  @ApiHeader({ name: 'keyspace', required: true })
+  createStandup(
+    @Body() body: StandupDTO,
+    @Headers('keyspace') keyspace: string,
+  ) {
+    return this.standupService.create(body, keyspace);
```
Cahllagerfeld, line 32: The provided Keyspace needs to be validated, as well as protected. So users cant get data which belong to a different server

eddiejaoude: Oh yes 👍 , it will be provided by the Discord Bot that we trust because it will have the authorised token. But we had everything for read (GETs) open which might be an issue

Cahllagerfeld (commented):
The complete process for me looks like this:
1. The Bot uses the token from the `env` to generate tokens for other clients that want to access the database
2. The bot itself could generate a token for itself when starting (here I could imagine adding a scope validation => The Bot gets writing scopes to the token, the other generated Tokens are only having a reading scope)
3. When communicating with the API we transer the JWT with `clientId, Keyspace, etc...`
4. We could potentially add an expiration time and stuff like that as well. 

By doing so we can check if the client that requests data is a `knwonCliet` for the requested keyspace and allow or reject the request

Cahllagerfeld:
Maybe we should discuss this in a voice channel as I think its a bit complex

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+export class AuthController {
+  constructor(private readonly authService: AuthService) {}
+
+  @Post()
+  @UseGuards(TokenGuard)
+  @ApiSecurity('token')
+  @ApiHeader({ name: 'keyspace', description: 'Keyspace' })
+  register(@Req() req) {
+    //TODO Hardcoded ServerID
+    return this.authService.register('123456', req.user.keyspace);
```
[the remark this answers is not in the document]
eddiejaoude: Thoughts/Questions:

- This is probably only allowed by the Discord server admins?
- It will be visible by everyone 🤔

eddiejaoude:
> Maybe we should discuss this in a voice channel as I think its a bit complex

wow this looks great 👍 yep let's go through it on a call to bounce ideas etc 🤓

src/auth/auth.controller.ts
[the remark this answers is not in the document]
Cahllagerfeld: exactly, I raised an issue for that in the Eddiebot-Repo already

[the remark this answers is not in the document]
eddiejaoude: ok great 👍

Cahllagerfeld:
Added Scopes and Scope Validation. Its all a bit messed up as its still work in progress, but it works in the scenario as I want it to work

src/auth/auth.service.ts
```diff
@@ -2,21 +2,21 @@ import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
...
+import { AuthDTO } from './dto/auth.dto';
 
 @Injectable()
 export class AuthService {
   //TODO move configCollection to database => own ConfigDataModule
   private configCollection: { [id: string]: { knownClients: string[] } } = {};
   constructor(private readonly JwtService: JwtService) {}
 
-  public register(serverId: string, keyspace: string): { accessToken: string } {
+  public register(body: AuthDTO): { accessToken: string } {
```
Cahllagerfeld, line 13: Data is transferred via a Body now

src/auth/guards/scopes.guard.ts
```diff
@@ -20,7 +20,8 @@ export class ScopesGuard implements CanActivate {
     return this.matchScopes(scopes, user);
   }
   private matchScopes(scopes: string[], user: any) {
-    if (user.scopes.some((scope) => scopes.includes(scope))) return true;
+    if (scopes.every((scope: string) => user.scopes.includes(scope)))
```
Cahllagerfeld, line 23: now every scope gets validated, not just some

src/auth/jwt.strategy.ts
```diff
@@ -14,8 +14,7 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
     super({
       jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
       ignoreExpiration: false,
-      //TODO: hardcoded Secret
-      secretOrKey: 'test',
+      secretOrKey: process.env.SECRET,
```
Cahllagerfeld, line 17: secret is read from `env`

src/astra/keyspace.interceptor.ts
```diff
@@ -0,0 +1,24 @@
...
+import { KeyspaceService } from './keyspace.service';
+
+@Injectable()
+export class KeyspaceInterceptor implements NestInterceptor {
+  private existingKeyspaces = [];
+  constructor(private readonly keyspaceService: KeyspaceService) {}
+  async intercept(context: ExecutionContext, next: CallHandler) {
+    const request: Request = context.switchToHttp().getRequest();
+    const user: TokenPayload = request.user as TokenPayload;
+    if (this.existingKeyspaces.includes(user.keyspace)) return next.handle();
```
Cahllagerfeld, line 18: currently handled in-memory. This could be put to database or retrieved with extra request

src/astra/keyspace.service.ts
```diff
@@ -0,0 +1,58 @@
...
+          username: this.config.get('STARGATE_USERNAME'),
+          password: this.config.get('STARGATE_PASSWORD'),
+        })
+        .toPromise();
+      return response.data.authToken;
+    }
+    throw new Error('Could not return AuthToken');
+  }
+
+  public async createKeyspace(serverId: string): Promise<void> {
```
Cahllagerfeld, line 44: function to create new keyspace, working for astra and local stargate

Cahllagerfeld (commented):
Added a keyspace-interceptor that creates a keyspace on the fly in case the keyspace isnt available in the database yet

src/astra/keyspace.service.ts
eddiejaoude: Great 👍

test/step-definitions/responses.ts
```diff
@@ -14,7 +22,20 @@ export class responses {
   public async before(): Promise<void> {
     const moduleFixture: TestingModule = await Test.createTestingModule({
       imports: [AppModule],
-    }).compile();
+    })
+      .overrideGuard(JWTGuard)
+      .useValue({
+        canActivate: (ctx: ExecutionContext) => {
```
Cahllagerfeld, line 28: I'm not quite sure about this: 
Should this be overwritten in an e2e-test?
Otherwise I would have to get a new token from the API, to decouple them.
Well I think thats in Theory no Problem, but this is the faster way. The Acquire-token-endpoint can be tested an a seperate test.

What do you think?

Cahllagerfeld (commented):
e2e Work 💥

test/step-definitions/responses.ts
eddiejaoude: I think with this in a `before`, all tests will have it, which will not be the case with some situation and we will not be able to test them.

I think as a separate step def then the test can use / not use to test both situation, plus if the test needs it can also pass parameters if needed

test/step-definitions/requests.ts
```diff
@@ -52,6 +53,27 @@ export class requests {
...
+  public async generateReadToken() {
+    const token = sign(
+      { scopes: ['Data.Read'], serverId: 'eddiehub' },
+      process.env.SECRET,
+    );
+    this.context.bearerToken = token;
+  }
+
+  @given(/authorization with Writing-Scopes/)
+  public async generateWriteToken() {
```
eddiejaoude, line 66: What do you think about merging these 2 functions into one?

From:
- `authorization with Reading-Scopes`
- `authorization with Writing-Scopes`
To:
- `authorization with "Reading-Scopes|Writing-Scopes"`

eddiejaoude:
Really amazing work 💥

test/step-definitions/requests.ts
Cahllagerfeld: yes I planned to do that, but didnt find time yet to implement it :D

eddiejaoude: ok great, no problem, maybe something we could do on a live stream? just an idea 🤓

Cahllagerfeld: sure, we could 👍

Cahllagerfeld: will we do this in a stream tomorrow then or sometime different

eddiejaoude: We could pair on this on tomorrow's live stream and get people involved, if you think it is a good time? Let me know me know what/when you prefer 🤓

PR #148, opened 2021-06-27
fix: tests for local+astradb (#138)

### LINKED ISSUE #138: read versionnumber directly from env
we could use `process.env.npm_package_version` instead of reading the version-number with the config-serivce and the env-variables like [here](https://github.com/EddieHubCommunity/api/blob/9813ac19d735da63fa93ceb36e0ffa74b3a76031/src/app.service.ts#L10).

### LINKED ISSUE #145: Run tests against local or remote
Ideally local is better, but some people cannot run Docker locally

### LINKED ISSUE #147: Error Handling for 500
There are several issues regarding error-handling.
These need to be found with failing e2e-tests and then handled by try-catch

eddiejaoude:
closes #147 
closes #145

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       stargate:
-        image: stargateio/stargate-3_11:v1.0.13
+        image: stargateio/stargate-3_11:v1.0.25
         ports:
           - 8080:8080
           - 8081:8081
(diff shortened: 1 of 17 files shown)


eddiejaoude:
@Cahllagerfeld I tried to put you as co-author on the last commit, as I used your patch. Not sure if it worked though

eddiejaoude:
Hmm those tests pass locally 🤔 @Cahllagerfeld could you run them locally and see what you get?

Cahllagerfeld:
> 
> 
> Hmm those tests pass locally 🤔 @Cahllagerfeld could you run them locally and see what you get?

Sure 👍

Cahllagerfeld:
There seem to fail 5 tests, thats the same amount like yesterday when we tested on your computer. I'll take a look at the logs as well🤔

eddiejaoude:
I think we got there 🎉 let me know what you think when you get time @Cahllagerfeld 🤓

Cahllagerfeld (approved):
I have 4 failing test, but all with `Error: function timed out, ensure the promise resolves within 5000 milliseconds` so it may be that my computer is under heavy load when running all tests with docker etc.. If they passed in the action and on your pc its fine I think

eddiejaoude:
Hm interesting. What happens if you test it against the cloud? just as a comparison

Cahllagerfeld:
> 
> 
> Hm interesting. What happens if you test it against the cloud? just as a comparison

all passed 🙏

eddiejaoude:
ok great 👍

PR #155, opened 2021-06-30
test: add e2e-test

### LINKED ISSUE #141: fix the commented test
fix the test : And the response should be "Welcome to EddieHub! Currently running version: v0.0.0" in : test/features/hello-world.feature

Cahllagerfeld:
closes #141

### DIFF
--- a/test/step-definitions/responses.ts
+++ b/test/step-definitions/responses.ts
@@ -164,11 +164,14 @@ export class responses {
     });
   }
 
-  @then(/the response should contain the Version/)
-  public validateAPIVersion() {
-    const versionString = this.context.response.text;
-    expect(versionString).to.match(
-      /Currently running version: \d{1,3}.\d{1,3}.\d{1,3}/,
-    );
+  @then(/the response-text should contain "([^"]*)"/)
+  public validateAPIVersion(text: string) {
+    const responseString = this.context.response.text;
+    if (/TYPE:/.test(text)) {
+      const regex = getRegex(text);
+      expect(responseString).to.match(regex);
+      return;
+    }
+    expect(responseString).to.equal(text);
   }
 }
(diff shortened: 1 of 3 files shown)


test/step-definitions/responses.ts
```diff
@@ -163,4 +163,12 @@ export class responses {
       }
     });
   }
+
+  @then(/the response should contain the Version/)
```
eddiejaoude, line 167: I think this should be more generic so we can reuse it

```suggestion
  @then(/the response should contain "([^"]*)"/)
```

Cahllagerfeld: And what should we put into the brackets when running the test? I think hardcoding a version-number wont work, because we have to adjust the test everytime then, and I'm not sure if I'm a fan putting a regex into the actual feature-file

eddiejaoude: Yep I agree, a regex in the feature file is not ideal

eddiejaoude (approved):
Looks great 👍

test/support/regexes.ts
```diff
@@ -4,6 +4,9 @@ const numberRegex = new RegExp(/[0-9]+/);
 const dateRegex = new RegExp(
   /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/,
 );
+const versionRegex = new RegExp(
+  /Currently running version: \d{1,3}.\d{1,3}.\d{1,3}/,
```
eddiejaoude, line 8: Is it possible to restrict this to just the version without the text?

Cahllagerfeld: sure, but then its not that strict any more

eddiejaoude: true, I guess it is finding a balance, I think it is fine for now

PR #168, opened 2021-07-23
Namespaces for data

### LINKED ISSUE #137: Using namespace for separate of data

eddiejaoude:
closes #137


Cahllagerfeld:
I added new test for the 403-scenarios. I think its mergeable then (I'd say as breaking change)

src/astra/astra-api.module.ts
```diff
@@ -0,0 +1,9 @@
+import { HttpModule, Module } from '@nestjs/common';
+import { KeyspaceService } from './keyspace.service';
+
+@Module({
+  imports: [HttpModule],
+  providers: [KeyspaceService],
+  exports: [KeyspaceService],
+})
+export class AstraApiModule {}
```
eddiejaoude, line 9: Maybe this could be called `Astra` to match the folder structure, then if it clashes in the imports we could use an alias

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+  @ApiSecurity('token')
+  @ApiQuery({ name: 'keyspace', required: true })
+  getTokens(@Query('keyspace') keyspace: string) {
+    return this.authService.getClientIds(keyspace);
+  }
+
+  @Post()
+  @UseGuards(TokenGuard)
+  @ApiSecurity('token')
+  // @ApiHeader({ name: 'keyspace', description: 'Keyspace', required: true })
```
eddiejaoude, line 32: Can this be removed?

src/astra/astra-api.module.ts
Cahllagerfeld: `AstraModule` already gets imported from my package, that's why I named it `AstraApi` but I think the alias should be no problem

src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,32 @@
...
+export const JWTGuard = AuthGuard('jwt');
+
+// export class MyAuthGuard extends AuthGuard(['jwt', 'discordGithub-strategy']) {
+//   constructor() {
+//     super();
+//   }
+//   canActivate(ctx: ExecutionContext) {
+//     return super.canActivate(ctx);
+//   }
+// }
```
eddiejaoude, line 32: Is this still needed?

test/step-definitions/requests.ts
```diff
@@ -75,11 +95,19 @@ export class requests {
...
+      post.set('Authorization', `Bearer ${this.context.bearerToken}`);
+    }
     this.context.response = await post.send(this.context.tableToObject(table));
 
-    this.context.preRequest = await request(
-      this.context.app.getHttpServer(),
-    ).get(url);
+    // this.context.preRequest = await request(
+    //   this.context.app.getHttpServer(),
+    // ).get(url);
```
eddiejaoude, line 105: Is this still needed?

eddiejaoude:
This looks amazing!! A few tiny comments. But really great work 💪 💥

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,48 @@
...
+@ApiTags('Auth')
+@Controller('auth')
+export class AuthController {
+  constructor(private readonly authService: AuthService) {}
+
+  @Get()
+  @UseGuards(TokenGuard)
+  @ApiSecurity('token')
+  @ApiQuery({ name: 'keyspace', required: true })
+  getTokens(@Query('keyspace') keyspace: string) {
```
eddiejaoude, line 25: I think this should be part of the path. I can change it in my PR

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,48 @@
...
+  @Delete()
+  @UseGuards(TokenGuard)
+  @ApiSecurity('token')
+  @ApiQuery({
+    name: 'token',
+    description: 'Token to delete',
+    required: true,
+  })
+  @HttpCode(204)
+  deleteClient(@Query() query) {
```
eddiejaoude, line 45: I think this should be part of the path, I can change it in my PR. What do you think?

test/step-definitions/requests.ts
Cahllagerfeld: I removed the comments

eddiejaoude: thanks 👍

Cahllagerfeld:
Is there something missing if we make the mentioned changes in the other PR?

eddiejaoude:
I think this is ready to go? any improvements we can make in the next PR

eddiejaoude:
I created the PR, so I cannot approve it 🤦‍♂️

Cahllagerfeld (approved):
Looks great 💥 
I'd merge it as a breaking change.

Cahllagerfeld:
> 
> 
> Looks great 💥
> I'd merge it as a breaking change.

@eddiejaoude 
One thing just came to my mind. You need to update your K8s secret, with the Token-Signing-Secret before deploying. Otherwise the token-creation will fail.

eddiejaoude:
> You need to update your K8s secret, with the Token-Signing-Secret before deploying

Oh good spot 👍 thank you, I will do it after dinner

eddiejaoude:
Creating the secret now on Kubernetes, but naming it `JWT_SECRET` so it is not confused with anything else - I will update the code afterwards

eddiejaoude:
JWT Secret added to Kubernetes 👇 

Screenshot 2021-07-31 at 09 33 09

PR #171, opened 2021-07-24
Move tokens from memory to DataStax DB

### LINKED ISSUE #165: Move tokens from memory to database
`​src/auth/auth.service.ts`

- [ ] rename variable to tokens
- [ ] use `tokens` collection

```ts
private configCollection: { [id: string]: { knownClients: string[] } } = {};
```

### LINKED ISSUE #173: create new endpoint on the auth module to validate tokens.

eddiejaoude:
closes #165 
closes #173 

- [x] write tests
- [ ] save data to db

Merge after #137c 

Cahl Todos

- [x] authService to database
- [x] Repair delete-token test


test/features/auth.feature
```diff
@@ -0,0 +1,38 @@
...
+            | scopes   | ["Data.Read"] |
+        Then the response status code should be 201
+        And the response should contain:
+            | clientId    | "TYPE:ID"     |
+            | keyspace    | "TYPE:STRING" |
+            | scopes      | ["Data.Read"] |
+            | accessToken | "TYPE:JWT"    |
+            | expiresIn   | "TYPE:NUMBER" |
+        When add bearer token to the header
+        # And restart app
```
eddiejaoude, line 36: This causes the test to fail until the tokens are persisted in the DB

Cahllagerfeld:
Do you want to merge this is branch 137c or do you want to merge this in main, when 137c is in main?

eddiejaoude:
we can do either, ideally into `main` when `137c` has been merged.

I think we have 2 options in preferred order:
1. if `137c` is already merged into `main`, then we can point this in `main`
2. if `137c` is not ready, we can merge this into `137c` it will just make it even bigger

src/auth/auth.service.ts
```diff
@@ -3,29 +3,38 @@ import { TokenPayload } from './interfaces/token-payload.interface';
...
+  ) {}
 
-  public register(body: AuthDTO) {
+  public async register(body: AuthDTO) {
     const clientId = uuidv4();
     const { serverId, scopes } = body;
+    let tokens: String[] = [];
+    try {
+      tokens = await this.astraService
+        .get<String[]>('Benjamin', serverId, 'tokens')
```
eddiejaoude: @Cahllagerfeld was `Benjamin` part of testing?

eddiejaoude: Oops I see you did another commit afterwards, may bad 🤦‍♂️ - just trying to work through conflicts

src/auth/auth.service.ts
```diff
@@ -3,63 +3,85 @@ import { TokenPayload } from './interfaces/token-payload.interface';
...
+        .replace('tokens', tokens, decoded.keyspace, 'tokens')
+        .toPromise();
     } catch (e) {
-      throw new HttpException('Invalid client id', HttpStatus.BAD_REQUEST);
+      throw new HttpException(
+        'Deleting Token went wrong',
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
     }
+    return;
```
eddiejaoude, line 85: I think we both code this function but you beat me to pushing it 😂 

I had this, it looks a bit different
```ts
public async removeClient(token: string) {
    if (!token)
      throw new HttpException('Please provide token', HttpStatus.BAD_REQUEST);

    const { clientId, keyspace } = this.jwtService.decode(
      token,
    ) as TokenPayload;

    try {
      await this.astraService.delete(clientId, keyspace, 'tokens');
    } catch (e) {
      throw new HttpException('Invalid client id', HttpStatus.BAD_REQUEST);
    }

    return;
  }
```

eddiejaoude: I can probably ditch my changes and port over the tests and validate endpoint

eddiejaoude:
We should merge `137c` branch into this PR

src/auth/auth.service.ts
eddiejaoude: I have brought over the tests I wrote. I think the delete needs a step def to use the ID

src/auth/auth.service.ts
```diff
@@ -3,63 +3,85 @@ import { TokenPayload } from './interfaces/token-payload.interface';
...
 
-  public register(body: AuthDTO) {
+  public async register(body: AuthDTO) {
     const clientId = uuidv4();
     const { serverId, scopes } = body;
+    let tokens: string[] = [];
+    try {
+      tokens = await this.astraService
+        .get<string[]>('tokens', serverId, 'tokens')
+        .toPromise();
```
eddiejaoude, line 22: Will this end point not create a new token every time? if so then can it added to the collection with a `create` rather than a `replace`?

src/auth/auth.service.ts
```diff
@@ -3,63 +3,85 @@ import { TokenPayload } from './interfaces/token-payload.interface';
...
+        .get<string[]>('tokens', decoded.keyspace, 'tokens')
+        .toPromise();
+    } catch (e) {
       return;
+    }
+
+    tokens = tokens.filter((clientID) => clientID !== decoded.clientId);
+    try {
+      await this.astraService
+        .replace('tokens', tokens, decoded.keyspace, 'tokens')
```
eddiejaoude, line 77: Can we not delete the document for that token with a delete ?

Cahllagerfeld:
I thin I did it wrong *facepalm*. I took a key `tokens` in the collection and stored an Array of strings (ids) in it. Its not generating strings inside the collection

eddiejaoude:
No problem, these things happen 🤓 

I think having each token and meta data saved in a document would be easier to manage, the DB can do the find/delete etc for us

src/auth/auth.service.ts
```diff
@@ -3,63 +3,85 @@ import { TokenPayload } from './interfaces/token-payload.interface';
...
+  public async validateClient(payload: TokenPayload): Promise<boolean> {
     const { keyspace, clientId } = payload;
-    if (
-      this.configCollection[keyspace] &&
-      this.configCollection[keyspace].knownClients.includes(clientId)
-    ) {
+    let tokens: string[];
+    try {
+      tokens = await this.astraService
+        .get<string[]>('tokens', keyspace, 'tokens')
```
Cahllagerfeld, line 48: store the tokens as keys in the collection, not within a function

Cahllagerfeld:
I'll take a look at this PR during the weekend, had quite much to do this week so I didnt find time for it :(

eddiejaoude:
No problem, that would be great 👍  I can look at it with you also

src/auth/auth.service.ts
```diff
@@ -25,47 +22,84 @@ export class AuthService {
...
-        decoded.keyspace
-      ].knownClients = this.configCollection[
-        decoded.keyspace
-      ].knownClients.filter((client) => client !== decoded.clientId);
-      console.log(this.configCollection);
-      return;
-    } catch (e) {
-      throw new HttpException('Invalid client id', HttpStatus.BAD_REQUEST);
+      this.jwtService.verify(token);
+      response.status(HttpStatus.OK).json({ valid: true });
```
Cahllagerfeld, line 100: For now I'm just validating if the token is valid. 

-  Do we have to do more checks here?
- How should the response look like at this endpoint? Is it just returning `{valid: true/false}` Or does it return more information?
- Does this endpoint needs to be secured, or can anybody post data to it for checking if it is valid?

Cahllagerfeld (commented):
Added some questions to it.

test/features/auth.feature
```diff
@@ -0,0 +1,97 @@
+@auth
+Feature: auth module
+
+    Scenario: get list of all tokens in invalid keyspace
```
Cahllagerfeld, line 4: added uncommented test

test/support/regexes.ts
```diff
@@ -4,9 +4,15 @@ const numberRegex = new RegExp(/[0-9]+/);
 const dateRegex = new RegExp(
   /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/,
 );
+const jwt = new RegExp(
+  /^[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*$/,
+);
 const versionRegex = new RegExp(
   /Currently running version: \d{1,3}.\d{1,3}.\d{1,3}/,
 );
+const v4 = new RegExp(
```
Cahllagerfeld, line 13: added regex for uuid

Cahllagerfeld (commented):
I think the PR is reviewable now. From my side its ready to merge

Cahllagerfeld:
We should add a test for checking if valid or invalid. Forgot about that lol

src/auth/auth.service.ts
Cahllagerfeld: Added the token-guard for using only with the bot to be consistent with all Auth-Endpoints

Cahllagerfeld:
Ready for merge in my opinion!! :D

src/auth/auth.service.ts
eddiejaoude: Ok sure 👍

eddiejaoude: > `response.status(HttpStatus.OK)`

I think it might make more sense to have the response code decision made in the controller and the service to just return the `boolean`

test/features/auth.feature
```diff
@@ -0,0 +1,97 @@
...
+            | clients | [] |
+
+    Scenario: get list of all tokens in my keyspace
+        Given authorisation
+        And make a POST request to "/auth/token" with:
+            | serverId | "eddiehub"    |
+            | scopes   | ["Data.Read"] |
+        When make a GET request to "/auth/token/eddiehub"
+        Then the response status code should be 200
+        And  the response property "clients" should be of type "TYPE:UUID"
```
eddiejaoude, line 18: Would this not be a collection of `UUID`s?

eddiejaoude:
Looking great 💥 , 2 small comments 🤓

test/features/auth.feature
Cahllagerfeld: In theory yes, but I return them [as an Array](https://github.com/EddieHubCommunity/api/pull/171/files#diff-aea62a4c305374791411369a46e8553d716a032a03bb48136176873887e3311cR91) because key and value are the same

src/auth/auth.service.ts
Cahllagerfeld: done

test/features/auth.feature
eddiejaoude: Oh I see now, sorry I did not check the step def. The step def checks for each item in the array, I read the feature test as it was only 1 item.

Maybe it could be renamed to something like...
```suggestion
        And the response property "clients" should have a collection of type "TYPE:UUID"
```

Cahllagerfeld (approved):
I think all open points are implemented now

eddiejaoude:
Awesome! Let's merge 🎉

Issue #161, opened 2021-07-20
Add AsciiDoc for Documentation?


Cahllagerfeld:
Should we add Ascii-Doc for Documentation (Auth-Flow, etc...) or is Swagger and standard `Readme` enough for this?

eddiejaoude:
Good question 🤔 

I think we might need AsciiDoctor to help write some proper documentation, that can then reference Swagger - we don't want to duplicate but them to complement each other

Cahllagerfeld:
I think this can be set up. The documentation itself is in #174.
The setup can be found in [this](https://github.com/eddiejaoude/stargate) repo

PR #179, opened 2021-08-06
fix: run docs for all changes (#161)

eddiejaoude:
Docs will run on all changes in `main`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,8 @@
 name: Docs
 on:
   push:
-    paths:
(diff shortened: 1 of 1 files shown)


Cahllagerfeld (approved):
In case the `path` doesnt work, I think thats the only possibility we have, or am I wrong? 🤔

Cahllagerfeld:
Maybe we can test it with #180 when this PR gets merged if the Action triggers with the `docs`-path

eddiejaoude:
Sure we can keep tweaking it

PR #186, opened 2021-08-10
fix: return 200 instead of 400

### LINKED ISSUE #185: [BUG] Not Return 400 with invalid token
### 🙁 Actual behavior

When [validating a token](https://github.com/EddieHubCommunity/api/blob/1b091078ca161410f970978f74da234fbebce34c/src/auth/auth.controller.ts#L59) a `400` gets returned in case the token is invalid.
I tried to implement this feature yesterday to the bot - Problem is when returning a 400 its treated like an exception which means that we need to catch it in the Bot.

### 🙂 Expected behaviour

In my opinion we should return `200` as the operation on the server was executed successfully, but only with the result, that the token is not valid

_No response_

Cahllagerfeld:
closes #185

### DIFF
--- a/src/auth/auth.controller.ts
+++ b/src/auth/auth.controller.ts
@@ -4,15 +4,12 @@ import {
   Delete,
   Get,
   HttpCode,
-  HttpStatus,
   Param,
   Post,
   Query,
-  Res,
   UseGuards,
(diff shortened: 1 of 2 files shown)


eddiejaoude (approved):
LGTM

PR #187, opened 2021-08-12
fix: when deleting dont validate expiration

### LINKED ISSUE #184: [BUG] Dont Validate expiration when delete
### 🙁 Actual behavior

When [deletig a token](https://github.com/EddieHubCommunity/api/blob/1b091078ca161410f970978f74da234fbebce34c/src/auth/auth.service.ts#L66) we verify the signature, that the Token or Client should be only deleted if the signature verifies. If validating the expiration we wont be able to delete expired tokens?

### 🙂 Expected behaviour

Is it okay, to not be able to delete expired tokens from the db? If no we should only validate the signature and not the expiration in addition

_No response_

Cahllagerfeld:
closes #184


eddiejaoude (approved):
LGTM

Cahllagerfeld:
First PR 100% made in the cloud ever 🔥

eddiejaoude:
oh nice 💥

PR #193, opened 2021-08-21
ci: adjust docker-action

Cahllagerfeld:


Cahllagerfeld:
Did we forget to setup node in this action?

Cahllagerfeld:
> 
> 
> Did we forget to setup node in this action?

Thats the only difference I see. You can have a look at the logs of the npm install in the failing action

Cahllagerfeld:
Thats the only difference I see. You can have a look at the logs of the failing action. I'm not sure if the action really installs the dependencies.

PR #199, opened 2021-10-08
feat: init keyspace interceptor

### LINKED ISSUE #198: [OTHER] Add Keyspace Interceptor for local Stargate
### What would you like to share?

It needs to be checked if the Api runs against a local stargate instance or the online Astra. When running against the local instance the keyspace interceptor can be used to create keyspaces on the fly, when using Astra it can't.

Todos

- [ ] Add Logic for checking if local stargate or not
- [ ] Add interceptor to relevant endpoints.

### Additional information

_No response_

Cahllagerfeld:
closes #198


eddiejaoude:
ah ok 👍

PR #208, opened 2021-10-09
feat: put to patch

### LINKED ISSUE #207: Turn `Put`-Endpoints to `Patch`
### What would you like to share?

What we are technically doing is a `Patch`, as we only update the properties with new ones. When doing patch, the whole object should be replaced.
Because of this reason we should turn our `Put`-Endpoints to `Patch`

### Additional information

_No response_

Cahllagerfeld:
closes #207

### DIFF
--- a/src/calendar/calendar.controller.ts
+++ b/src/calendar/calendar.controller.ts
@@ -6,7 +6,8 @@ import {
   HttpCode,
   Param,
   Patch,
-  Post, UseGuards
+  Post,
+  UseGuards,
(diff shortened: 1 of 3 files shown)


eddiejaoude:
Great improvements 👍 

In the service only the sent fields are updated, so these should be `patch` 👍 https://github.com/EddieHubCommunity/api/blob/d82a08794cb6d3412945fa53d221da4852a29ede/src/calendar/calendar.service.ts#L121

Issue #215, opened 2021-10-20
Change License copyright to `Eddie Jaoude`


schmelto:
Thanks everyone 👍 Makes sense. There is a mix across our Org repos, they should then be all updated to be `Eddie Jaoude`

_Originally posted by @eddiejaoude in https://github.com/EddieHubCommunity/awesome-github-profiles/issues/712#issuecomment-947703086_

aroramrinaal:
@Cahllagerfeld , Can I work on this one?

Cahllagerfeld:
Sure 👍

aroramrinaal:
@Cahllagerfeld , I just opened a pull request for this #247

eddiejaoude:
Thanks 👍

PR #216, opened 2021-10-20
Update LICENSE

rushikeshadhav:
#215

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/api/pull/216"></a>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Eddie Jaoude Community
+Copyright (c) 2020 Eddie Jaoude 
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Eddie Jaoude Community
+Copyright (c) 2020 Eddie Jaoude
```
krishguptadev, line 3:
```suggestion
Copyright (c) 2020 EddieHubCommunity 
```

krishguptadev (changes requested):

LICENSE
Pradumnasaraf: No, 
I think it should be ***` Eddie Jaoude`***, and also there is a previous conversation regarding that.
You can also check LinkFree Licence - [`Licence`](https://github.com/EddieHubCommunity/LinkFree/blob/6d3b1ce272709bafea85b40cd8a797118751c695/LICENSE#L3).

@SaraJaoude will explain it with more clarity.
cc - @eddiejaoude

krishguptadev: Ohk!

krishguptadev (approved):

PR #232, opened 2021-11-06
Add Standup Module

### LINKED ISSUE #221: Migrate Standup-Module to MongoDB

Cahllagerfeld:
closes #221 

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/api/pull/232"></a>


Cahllagerfeld:
All in all more deleted lines than new ones. Wouldnt have expected that.

src/standup/standup.controller.ts
```diff
@@ -30,26 +23,33 @@ export class StandupController {
...
-  createStandup(@Body() body: StandupDTO) {
-    return body;
+  async createStandup(@Body() body: StandupDTO) {
+    return this.standupService.createStandup(body);
   }
 
   @Get()
-  findAllStandups(@User() user) {}
+  async findAllStandups() {
+    return await this.standupService.getAllStandups();
```
eddiejaoude, line 32: I wonder if the controller and service methods should be the same when doing the same thing? For example both could be `findAll`, it is in the Standup controller / service already, so not sure that name needs appending

Cahllagerfeld: updated it

eddiejaoude: Thanks 👍 

Sorry to be a pain, do you think the appended `Standups` is needed? When these are in the class names already, so when we use it is currently `this.standupService.getAllStandups()` but could be shortened to `this.standupService.getAll()` (I prefer `find` tbh but not a biggie)

Cahllagerfeld: I removed the appended Stanup as mentioned. In the controller function-name I'd leave it in to be honest so its clearer also when not seeing the class name, maybe because its scrolled out.

eddiejaoude (approved):
Looks good 👍

src/standup/standup.controller.ts
eddiejaoude: Sure 👍

PR #253, opened 2022-03-19
Github Events Module

Cahllagerfeld:
- [x] write Module for Events
- [x] Adjust Docs
- [x] add option for gh-token & rate-limit

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/api/pull/253"></a>


src/github/dto/create-github.dto.ts
```diff
@@ -1,14 +1,9 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsNotEmpty, IsString } from 'class-validator';
+import { IsNotEmpty, IsOptional, IsString } from 'class-validator';
```
krshdev, line 2: I don't know if this is necessary but I think the linter is right, and we should remove this 😅 

```suggestion
import { IsNotEmpty, IsString } from 'class-validator';
```

PR #256, opened 2022-03-23
Store Ids in Lowercase

Cahllagerfeld:


Cahllagerfeld:
Github doesn't transform Path Params to lowercase, so it seems like not being a good practise

PR #268, opened 2022-05-03
feat: add throttling

Cahllagerfeld:


src/app.module.ts
```diff
@@ -7,9 +7,15 @@ import { MongooseModule } from '@nestjs/mongoose';
...
 import { UserModule } from './user/user.module';
 import { GithubModule } from './github/github.module';
+import { ThrottlerGuard, ThrottlerModule } from '@nestjs/throttler';
+import { APP_GUARD } from '@nestjs/core';
 
 @Module({
   imports: [
+    ThrottlerModule.forRoot({
+      ttl: 60,
+      limit: 60,
```
Cahllagerfeld, line 17: Added a rate-limit which is 60 Requests per Minute from the same IP. In other words like 1 request per second. We should try out how this is going, and adjust in the future

PR #277, opened 2022-06-02
feat: remove events if user deletes

Cahllagerfeld:


Cahllagerfeld (commented):
@eddiejaoude quick question: Given the case a user opts out and deletes their github account from the API, would we delete their detail-information about their Events as well? When speaking about events I mean the events that get stored as dedicated Objects in the DB. The bumped numbers in the Eddiehub-Profile would remain untouched.

I personally have the opinion we should delete all Data that is related to that account and thats also the purpose of the PR.

eddiejaoude (approved):
Looks good 👍

eddiejaoude:
> I personally have the opinion we should delete all Data that is related to that account 
> in case they decide afterwards to re opt-in, everything gets tracked properly again

I think this is the best way too 👍

PR #289, opened 2022-07-01
feat: add pagination to endpoint

### LINKED ISSUE #288: Introduce Pagination
### 🙁 Actual behavior

Our Response take quite long, due to the huge payloads, when displaying all records

### 🙂 Expected behaviour

We should add pagination to make the transferred payload more lightweight

_No response_

Cahllagerfeld:
closes #288

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/api/pull/289"></a>

### DIFF
--- a/test/step-definitions/responses.ts
+++ b/test/step-definitions/responses.ts
@@ -108,4 +108,38 @@ export class responses {
       }
     });
   }
+
+  @then(/the reponse in property "([^"]*)" at index "([^"]*)" should contain:/)
+  public validateResponseProperty(
+    property: string,
+    index: string,
+    table: { rawTable: [] },
+  ) {
+    const data = this.context.tableToObject(table);
+    Object.keys(data).forEach((key) => {
+      if (/TYPE:/.test(data[key])) {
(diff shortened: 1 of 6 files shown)


Cahllagerfeld (commented):
We prob need to add that to other endpoints as well

test/step-definitions/responses.ts
```diff
@@ -108,4 +108,38 @@ export class responses {
...
+            expect(isValidID).to.be.true;
+            break;
+          default:
+            const regex = getRegex(data[key]);
+            expect(
+              JSON.parse(this.context.response.text)[property][index][key],
+            ).to.match(regex);
+            break;
+        }
+      } else {
```
eddiejaoude, line 138: I find it easier to read if the  `if` ends with a `return` and then the `else` is not required

Cahllagerfeld: You got me there :D this test was generated by Copilot xD I'll refactor it 👍

eddiejaoude (approved):
LGTM 👍

Issue #292, opened 2022-07-04
[OTHER] redundant return await affects performance  


jameslieu:
### What would you like to share?

I've seen a few times `return await` is used outside try/catch blocks. This introduces a unnecessary microtask which impacts performance a little. 

This link explains this in some more detail
https://eslint.org/docs/latest/rules/no-return-await 

If you don't mind, I'd like to contribute and open a PR to remove `return await` from areas where they're used outside a try/catch block. Thanks

jameslieu:
I've opened a PR for this but it was subsequently closed. I'll also close this issue as well

eddiejaoude:
Thank you James!

We will geek out on some other issues/PR soon!

PR #293, opened 2022-07-04
fix: remove 'return await' to improve performance

jameslieu:
## Fixes Issue
Closes https://github.com/EddieHubCommunity/api/issues/292

## Check List (Check all the applicable boxes) 

- [x] My code follows the code style of this project.
- [x] All new and existing tests passed.
- [x] This PR does not contain plagiarized content.
- [x] The title of my pull request is a short description of the requested changes.

<a href="https://gitpod.io/#https://github.com/EddieHubCommunity/api/pull/293"></a>

### DIFF
--- a/src/github/github-event.service.ts
+++ b/src/github/github-event.service.ts
@@ -38,7 +38,8 @@ export class GithubEventService {
       return createdEvent.populate('githubUsername', '-__v');
     }
 
-    return newEvent.save();
+    await newEvent.save();
+    return;
   }
 
   public async getAll() {
--- a/src/github/github-event.service.ts
+++ b/src/github/github-event.service.ts
@@ -35,25 +35,25 @@ export class GithubEventService {
 
     if (populate) {
       const createdEvent = await newEvent.save();
-      return await createdEvent.populate('githubUsername', '-__v');
+      return createdEvent.populate('githubUsername', '-__v');
     }
 
-    return await newEvent.save();
+    return newEvent.save();
   }
 
   public async getAll() {
-    return await this.eventModel.find();
+    return this.eventModel.find();
   }
 
   public async getOne(id: string) {
-    return await this.eventModel.findById(id);
+    return this.eventModel.findById(id);
   }
 
   public async getByUsername(githubUsername: string) {
-    return await this.eventModel.find({ githubUsername });
+    return this.eventModel.find({ githubUsername });
   }
 
   public async deleteByUsername(githubUsername: string) {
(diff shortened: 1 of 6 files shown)


src/github/github-event.service.ts
```diff
@@ -35,25 +35,25 @@ export class GithubEventService {
 
     if (populate) {
       const createdEvent = await newEvent.save();
-      return await createdEvent.populate('githubUsername', '-__v');
+      return createdEvent.populate('githubUsername', '-__v');
     }
 
-    return await newEvent.save();
+    return newEvent.save();
```
eddiejaoude, line 41: I think the `await` is important otherwise the API will return a `200` but the data might not have been saved yet

jameslieu: Thanks for your feedback. I've made an update. Please have a look when you have a moment

eddiejaoude: Thanks James! But I think the same applies to all the `await`s that were removed, I think they are all still needed

eddiejaoude: Btw I was just reading the link you shared in issue #292 , I would be interested to get @Cahllagerfeld thoughts on it

jameslieu: Sure, the thing with `return await` is that it is redundant outside of a try/catch block and is considered by some people a bad practice. You have tests which should alieviate your concerns. 

Of course it doesn't necessarily "break" your code, it does have a slight performance impact, that said, I'm also happy for this PR to be closed also 👍  

But I'll wait to see @Cahllagerfeld 's thoughts

eddiejaoude: Thanks James for explaining 👍 . I was not aware of this performance issue, so I need to do more research to understand better.

Plus I feel you might have highlighted something else, maybe we should be wrapping these in `try/catch` 🤔  (whatever happens, I feel I have/will learn something new, so thank you 🔥 )

Cahllagerfeld: I never heard of this before either, so I can't give a statement on this... The errors are handled inside the Services, so there won't be anything to catch on Controller-Level (at least I hope so). I didn't check this issue in detail yet, would also need to do some research on it.

eddiejaoude:
This is really interesting James, I did not know this!

https://eslint.org/docs/latest/rules/no-return-await
> You can avoid the extra microtask by not awaiting the return value, with the trade off of the function no longer being a part of the stack trace if an error is thrown asynchronously from the Promise being returned. This can make debugging more difficult.

My concern is, it makes debugging more difficult, which I would prefer to avoid

